PR #176, opened 2026-05-11
Add Helm chart for deployment in Kubernetes

jwoglom:
Adds a Kubernetes Helm chart which I was able to use to successfully stand up a basic Nocturne install via ArgoCD.

This isn’t battle tested (i am still working through setup steps in my homelab) but, given that production deployments of Nocturne may very well use Kubernetes, it seems like a good starting point to get an opinionated helm chart into the repo.

image


ryceg:
Out on a walk so I can't really review it right now but I figured that you're probably the best person to actually have a look at the latest release of aspire, which has added a lot of support for kubernetes things- as the domain expert, does the aspire k8s API cover our needs?

https://aspire.dev/whats-new/aspire-13-3/

jwoglom:
> Out on a walk so I can't really review it right now but I figured that you're probably the best person to actually have a look at the latest release of aspire, which has added a lot of support for kubernetes things- as the domain expert, does the aspire k8s API cover our needs?
> 
> https://aspire.dev/whats-new/aspire-13-3/

Hm, I’ll take a look and report back!

jwoglom:
@ryceg I asked Claude to analyze upstream Aspire's kubernetes support vs. this custom chart:
> TL;DR
                                                                                                                                                                                                      Stay on the hand-written chart for now. Aspire's Helm publishing fits greenfield apps with simple topologies, but doesn't yet emit the primitives your chart depends on (Job, Ingress, HPA, PDB, initContainers, digest pinning, securityContext hardening).
The current chart is doing work Aspire's publisher can't yet do, and what it can do (env wiring, connection-string stitching) is the cheapest      part to maintain by hand. At v0.1.0-alpha with RLS-enforced multitenancy, the audit clarity of an explicit hand-written chart outweighs regeneration     convenience.
                                                                                                                                             Re-evaluate when: (a) Aspire emits Job and Ingress natively, (b) ASPIRECOMPUTE003 ships GA with digest support, and (c) a second deployable surface      makes the hand-written boilerplate cost painful. Until then, the chart is an asset, not tech debt.


While we _could_ use aspire's builtin support it seems that it wouldn't be as native and we would have to hack around it and still have custom additions on top of it. Seems reasonable to re-evaluate on future major aspire releases.

PR #201, opened 2026-05-12
fix(charts): prevent each_key_duplicate crashes from duplicate-timestamp keys

MrJarnould:
## Summary

Fixes a class of `each_key_duplicate` crashes that take down the dashboard error boundary when the source data has multiple items sharing the same timestamp. Three components were affected:

- **Glucose chart merge** — `chart-data-engine.svelte.ts` could produce duplicate-mills points when realtime entries overlapped server data; `GlucoseTrack.svelte` used a Date object as the keyed-each key.
- **Device event markers** — keyed by `marker.time.getTime()` alone. When multiple events share a timestamp (e.g., a connector sync that produces several SiteChange events at the same instant), the reconciler throws.
- **Actogram** — keyed by `dataRow.day.getTime()` alone. Defensive fix; same pattern.

I personally hit this in production with 300+ SiteChange device events all carrying the same timestamp from a single sync batch, which collapsed the entire dashboard into the error overlay. Confirmed via Brave DevTools that the duplicate key was `1778537825137` (a `getTime()` value) and the `Proxy(Array)` being iterated held identical-timestamp markers.

## Root cause

Svelte 5's keyed each enforces unique keys via `SameValueZero`. Several components on the dashboard used `time.getTime()` as the key with no tiebreaker, assuming timestamp uniqueness. That assumption fails when the upstream data has duplicates — which can happen from realtime/server merge overlap, connector backfills, or import scripts.

When the key collides, the error is thrown during reactive flush (not during user-code execution), so the stack trace points only at Svelte runtime chunks and gives no direct hint about which component is failing. Identification required setting a breakpoint inside the error function and inspecting the iterated array's shape.

### `chart-data-engine.svelte.ts` (glucose merge)

Replaced the filter+merge approach with a `Map<mills, GlucosePoint>` that dedupes by mills. Server data wins on collision; realtime fills gaps. This is correct under all four input shapes: duplicates in base, duplicates in realtime, overlap between base and realtime, no duplicates at all. Previous code only checked realtime against base, leaving two holes (base-vs-base and realtime-vs-realtime).

### `GlucoseTrack.svelte`

Changed key from `(point.time)` (a Date object — reference-compared) to `(point.time.getTime())` (a primitive number — value-compared). Matches the pattern already used in `DeviceEventMarkers`.

### `DeviceEventMarkers.svelte`

Changed key from `(marker.time.getTime())` to `(marker.treatmentId ?? \`\${marker.time.getTime()}-\${i}\`)`. Prefers the GUID treatmentId when available; falls back to a timestamp+index combination for markers without an ID.

### `Actogram.svelte`

Changed key from `(dataRow.day.getTime())` to `(\`\${dataRow.day.getTime()}-\${i}\`)`. Defensive — same pattern.

## Test plan

- [ ] Dashboard renders without `each_key_duplicate` when realtime overlaps server data
- [ ] Dashboard renders without `each_key_duplicate` when the device-event marker array contains duplicate timestamps
- [ ] Glucose chart still updates in real time
- [ ] Server-side glucose points take priority over realtime when both exist for the same mills
- [ ] Actogram renders without `each_key_duplicate` even if the same day appears twice in the data

## Verification done

Built and deployed the branch image to a production multi-tenant deployment with the bug reproducing reliably (300+ same-timestamp SiteChange events from a Dexcom-side data anomaly). After the fix, the dashboard, `/meals`, `/calendar`, and other previously-failing routes render cleanly.

### DIFF
--- a/src/Web/packages/app/src/lib/components/actogram/Actogram.svelte
+++ b/src/Web/packages/app/src/lib/components/actogram/Actogram.svelte
@@ -125,7 +125,7 @@
   </div>
 
   <!-- Rows -->
-  {#each visibleDataRows as dataRow, i (dataRow.day.getTime())}
+  {#each visibleDataRows as dataRow, i (`${dataRow.day.getTime()}-${i}`)}
     <div
       class="flex items-center"
       style:height="{rowHeight}px"
--- a/src/Web/packages/app/src/lib/components/dashboard/glucose-chart/engine/chart-data-engine.svelte.ts
+++ b/src/Web/packages/app/src/lib/components/dashboard/glucose-chart/engine/chart-data-engine.svelte.ts
@@ -486,6 +486,10 @@ export function createChartDataEngine(
   });
 
   // ---- Glucose data (merged with realtime) ----
+  // Dedupe by timestamp: server data wins on collision, realtime fills gaps.
+  // A keyed {#each} downstream requires a unique key per point, so the same
+  // mills value must never appear twice — even if base or realtimeStore
+  // emit duplicates.
   const glucoseData = $derived.by(() => {
     const base = serverChartData?.glucoseData ?? [];
     if (!serverChartData) return base as GlucosePoint[];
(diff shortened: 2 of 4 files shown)


MrJarnould:
Superseded by #202 — reopened from a more descriptive branch name (`fix/charts-each-key-duplicate`) that covers all three keyed-each fixes on the branch, not just the original glucose-chart one.

PR #234, opened 2026-05-16
fix(di): tolerate missing Nightscout write-back sinks

MrJarnould:
## Summary

- Restores `GET /api/v4/glucose/sensor` and the legacy `POST /api/v1/{entries,treatments,devicestatus}.json` endpoints, which were returning 500 in production
- Root failure surfaced as `Unable to resolve service for type 'NightscoutConnectorConfiguration' while attempting to activate 'NightscoutTreatmentWriteBackSink'` — the typed HttpClient factory couldn't see the connector config singleton even though the Nightscout BG service had it injected at startup (so the singleton clearly exists; the per-request scope just can't reach it). The deeper DI registration bug is still open for follow-up; this PR stops it from cascading into a request-pipeline outage.
- The `IDataEventSink<T>` factories for Treatment / Entry / DeviceStatus / Profile / Food / Activity now resolve their Nightscout write-back sink through a new `TryAddOptionalSink` helper. If construction throws, it logs a warning and skips the sink instead of bubbling up. SignalR sinks (essential for realtime) stay required.

## Impact

- v4 dashboard reads (`/api/v4/glucose/sensor`) start working again, unblocking the BG Delta widget and the realtime store backfill
- Trio (and any other Nightscout-compat uploader) can POST to v1 endpoints again
- Side effect: Nightscout write-back becomes a no-op when the underlying registration is broken. For self-hosted Nocturne instances with no upstream Nightscout to mirror to, this is the correct behavior. For deployments that actually rely on write-back, the warning log will surface the underlying registration problem.

## Test plan

- [x] `dotnet build src/API/Nocturne.API/Nocturne.API.csproj` — 0 errors
- [x] `dotnet test tests/Unit/Nocturne.API.Tests` — 3104 passed, 0 failed
- [ ] After deploy: `curl https://nocturne.jacob-arnould.com/api/v4/glucose/sensor?limit=10` returns 200
- [ ] After deploy: Trio "Connect to Nightscout" succeeds
- [ ] After deploy: API logs show `Optional NightscoutTreatmentWriteBackSink could not be resolved; skipping` warnings (confirms the underlying DI bug is still present and surfaced)

## Follow-up

Open question worth a separate investigation: why is `NightscoutConnectorConfiguration` (singleton, registered in `NightscoutConnectorInstaller.Install` and provably available — the BG service gets it injected) not visible to `DefaultTypedHttpClientFactory<T>` from a request scope? Same pattern affects every connector (Glooko/CareLink/etc.), so it's systemic to `AddConnectorConfiguration`.

### DIFF
--- a/src/API/Nocturne.API/Extensions/ServiceRegistrationExtensions.cs
+++ b/src/API/Nocturne.API/Extensions/ServiceRegistrationExtensions.cs
@@ -394,12 +394,11 @@ public static class ServiceRegistrationExtensions
         services.AddScoped<ITreatmentCache, Nocturne.API.Services.Treatments.TreatmentCacheAdapter>();
         services.AddScoped<SignalRTreatmentEventSink>();
         services.AddScoped<IDataEventSink<Treatment>>(sp =>
-            new CompositeDataEventSink<Treatment>(
-                [
-                    sp.GetRequiredService<SignalRTreatmentEventSink>(),
-                    sp.GetRequiredService<NightscoutTreatmentWriteBackSink>()
-                ],
-                sp.GetService<ILogger<CompositeDataEventSink<Treatment>>>()));
+        {
+            var sinks = new List<IDataEventSink<Treatment>> { sp.GetRequiredService<SignalRTreatmentEventSink>() };
+            sp.TryAddOptionalSink<NightscoutTreatmentWriteBackSink, Treatment>(sinks);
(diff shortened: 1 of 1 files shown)


ryceg:
Thanks for identifying this- my bad for giving Claude a long leash last night. This is superseded by #236 which should fix the problem once and for all

PR #239, opened 2026-05-17
Change default alarm title/favicon flash duration from 500ms to 1s

jwoglom:
The previous 500ms cadence made the tab title and favicon flicker twice per second during HIGH/LOW alarms, which was IMO too aggressive as a default given that it seems to occur regardless of whether the tab is focused or not. Let me know if you disagree.


ryceg:
Yeah that's a good call

jwoglom:
fyi this is how fast it looked on Safari on an iPad

https://github.com/user-attachments/assets/4a13c116-7450-46e6-8d4d-24901f220bd7

Issue #251, opened 2026-05-23
Inventory tracking — architecture & phased delivery plan


ryceg:
# Inventory tracking — architecture & phased delivery plan

**This issue is the architectural spec.** Six implementation issues link back to it as `Part of #251`. Read this first before opening any inventory PR. Update this issue's body — not via PR — as decisions evolve.

## Background

A prior attempt at this feature landed on `MrJarnould/nocturne#feat/inventory-tracking`. The inventory engine (FEFO consumption, immutable ledger, idempotency, reversal, RLS) is sound — that work will be lifted. The catalog model and the integration with `PatientDevice` / `PatientInsulin` are not, and that PR will not be merged. This spec resets the architecture before any new code lands.

---

### Four-level data model

| Level | What it represents | Entity |
|---|---|---|
| Model | "Omnipod 5 system", "Dexcom G7" — what the patient *owns* | `DeviceCatalogEntry`, `InsulinFormulation` |
| SKU | "Omnipod 5 pods", "Humalog 3ml KwikPen" — what the patient *stocks* | `ConsumableCatalogEntry`, `InsulinSku` |
| Batch | "Box of 5 pods, lot ABC, exp 2027-06" — what the patient *bought* | `InventoryBatchEntity` |
| Instance | "Pod #4 from that box, left abdomen, 2026-05-22 → 2026-05-25" — what the patient *used* | `ConsumableInstanceEntity` |

### Durables are not consumables

A device the patient **owns and reuses** is a `PatientDevice`, not inventory. A thing the patient **buys, uses, and discards/depletes** is a consumable, tracked in inventory. The boundary matters most for transmitters:

- **Dexcom G6 transmitter** — disposable, fixed 90-day life, bought and thrown away. A consumable: gets a `ConsumableCatalogEntry`, has `WearDays`, inventory-tracked.
- **Eversense transmitter** — rechargeable, reusable across sensor insertions, multi-year life, recharged nightly. A durable: **no** `ConsumableCatalogEntry`, **no** `WearDays`, **not** inventory-tracked. Only the implanted sensor (180d / 365d) is the consumable.

Battery state is **never** modelled as inventory lifespan:
- Battery charge % is runtime telemetry (changes hourly) — a separate device-health concern, out of scope here.
- Disposable batteries (e.g. Medtronic AAA) are a `GenericSupplyKind.PumpBattery` consumed at a variable rate — covered by `HistoricalRate` projection, no special model.
[...]

PR #286, opened 2026-06-01
feat(inventory): add consumable wear sessions with DeviceEvent hook (Phase 1)

MrJarnould:
Part of #251 — Phase 1 of the phased inventory rollout.

> **Dependency:** this branch sits on top of Phase 0 (catalog composition + ConsumableCatalog + InsulinSku). The Phase 0 commit (`b820ec46`) is included in the diff and the Phase 0 PR should land first; once that merges, this branch's diff will narrow to just the Phase 1 changes.

## Summary
Introduces `ConsumableInstance` — a per-tenant record of a physical CGM sensor / pod / infusion set being worn. Wear sessions open and close automatically in response to `DeviceEvent`s.

- **`SensorStart`** → opens a `CgmSensor` instance, resolving brand from the patient's current CGM `PatientDevice`
- **`SiteChange`** → opens a `Pod` (tubeless pump) or `InfusionSet` (tubed pump) based on `PumpCapability.IsTubeless` on the catalog entry
- A new event closes the prior open of the same kind with `EndReason.Planned`
- **Idempotent**: each instance carries `SourceDeviceEventId`, with a unique partial index — resyncs and reprocessing are safe

## Where it hooks in
- `DeviceEventController` (`Create` + `Update`) — live single-event path
- `TreatmentDecomposer.DecomposeDeviceEventAsync` — single-event projection path from legacy treatments
- A 30 s-delayed `BackgroundService` per-tenant backfill, gated by `AnyExistAsync`, replays historical `SensorStart` / `SiteChange` events for existing tenants

## API surface
`/api/v4/consumable-instances`:
- `GET /active` — open sessions for the current tenant
- `GET /recent?limit=` — most recent closed sessions across all kinds
- `GET /{id}`
- `PATCH /{id}` — editable subset: insertion site, serial, notes, ended-at, end reason, residual units
- `DELETE /{id}` — soft delete

## Frontend
New **Wear sessions** card on `/settings/patient` (`ConsumableInstanceHistory.svelte`):
- Active card per kind (CGM sensor / pod / infusion set) with started-at, worn duration vs catalog wear-days, and insertion site
- Recent list with end reason badge
- Edit / End-now dialog driven by the `update` remote function
- Soft-delete with confirm dialog

## Storage
- `consumable_instances` table with **ENABLE + FORCE ROW LEVEL SECURITY** and a `tenant_isolation` policy (USING + WITH CHECK)
- Indexes: `(tenant_id, kind, ended_at)` for the open-by-kind lookup, unique partial index on `source_device_event_id` for the idempotency guard
- `tenant_id` FK to `tenants(id)` with `ON DELETE CASCADE`
- `InventoryItemId` / `InventoryBatchId` columns nullable; populated in Phase 2 / Phase 4 once those tables exist

RLS coverage is picked up automatically by `RlsCompletenessFixture` via the `ITenantScoped` marker on the entity — no exclusion-list edits.

## Test plan
- [x] `dotnet build` clean across Models, Infrastructure, API, Tests
- [x] `dotnet test` non-Integration: 3146 passed (1 pre-existing E2E failure unrelated to this PR — Aspire `mermaid-publish` / `publish-compose` pipeline ordering)
- [x] 8 new `ConsumableInstanceServiceTests`: idempotency, unsupported-type no-op, sensor open, prior-open close, no-CGM skip, tubeless → Pod, tubed → InfusionSet, update-patch
- [x] `pnpm run check` — no new errors introduced (baseline 811 → 808 after fixes in this branch; remaining 5 in the new generated remote module mirror an existing template pattern repeated 413× across all `*.generated.remote.ts` files)
- [ ] Manual smoke once Phase 0 lands: create a SensorStart via the API, verify a CgmSensor instance opens and shows in `/settings/patient`; create a second one, verify the first closes with `EndReason.Planned`
- [ ] RLS: confirm `RlsCompletenessFixture` passes on CI


MrJarnould:
Splitting off one question from a UX review @ETolboom shared ([MrJarnould/nocturne#1](https://github.com/MrJarnould/nocturne/pull/1)) since it's specific to Phase 1's design:

The current hook uses a strict allowlist:
- `SensorStart` → `CgmSensor`
- `SiteChange` → `Pod` (tubeless) or `InfusionSet` (tubed)

ETolboom raised the case where different uploaders emit different event types for the same physical action — e.g. `PodChange` instead of `SiteChange` from some sources. His suggestion was configurable mapping (multi-select per consumable kind).

My current thinking: keep Phase 1 strict so we don't ship a flexible-mapping surface before we have real-world data on what uploaders actually emit, then revisit in Phase 5 when we know which event types are showing up in the wild. The cost of being wrong is low — adding event types to the allowlist later is additive and backfill is idempotent.

Is that policy in line with what you'd want, or would you rather see configurable mapping land in Phase 1?

Broader UX direction from his review is over in this Discussion: https://github.com/nightscout/nocturne/discussions/288

ETolboom:
> My current thinking: keep Phase 1 strict so we don't ship a flexible-mapping surface before we have real-world data on what uploaders actually emit, then revisit in Phase 5 when we know which event types are showing up in the wild. The cost of being wrong is low — adding event types to the allowlist later is additive and backfill is idempotent.

My thinking was that the uploaders get mapped to an internal (domain) model. This makes sure that all possible event types are known and, by extension, are normalized. Thats why I think its not much of an issue. 

Whether a ``SiteChange`` is specifcally for a tubed system or pod system should be irrelevant. A simple matter of checking whether  ``InfusionSet`` inherits from ``SiteChange`` should suffice. In other words, I think we can make use of the fact that we use C# and we have a proper type system.

PR #289, opened 2026-06-01
Adding manual insulin importing from Glooko; Adding code to create a patientInsulin record during import if none matches the insulin treatment being imported

awilner:


ryceg:
This looks like a solid start, but I think it would benefit from another pass. Sync identifier is for client sourced ids (think: Trio, etc, who already have their own method of identifying records), but you're missing not implementing the legacyid which conceptually is Nightscout API v3's identifier. It's the ID a device/uploader assigns and reuses so the same logical record can be matched and updated across systems. When a source gives you a real stable ID, you put it here; the controller's idempotency check keys on (data_source, sync_identifier).

The rough split: LegacyId = "where this came from / a hash of its content," SyncIdentifier = "the external system's own stable handle for this record."

I realise that the distinction between the two isn't super clear, I'll write up some docs foor inclusion in the scalar. I can also have Claude take a squiz at it to just double check stuff.

ryceg (commented):
## Review — manual insulin importing from Glooko

Nice feature. The wiring is sound — DI registrations, repos, the interface, the enum, and the test constructor all line up, and the old `TODO` is properly resolved. I found one bug I'd treat as a blocker, plus a few smaller things.

### 🔴 High — Basal injection re-sync is not idempotent

`MapV3ManualInsulin` sets `LegacyId` on each `BasalInjection` but never sets `SyncIdentifier`. The persistence paths don't line up with that choice:

- **Boluses** are fine: `BolusRepository.BulkCreateAsync` has a DB-level dedup pass keyed on `LegacyId` (`BolusRepository.cs:409-427`), so the manual *boluses* dedupe correctly on re-sync.
- **Basal injections** go through `BasalInjectionRepository.CreateAsync`, whose upsert only fires when `SyncIdentifier` is non-empty (`BasalInjectionRepository.cs:107-119`). With `SyncIdentifier == null` it skips dedup and goes straight to `Add` + `SaveChanges`.

Meanwhile the table has a unique partial index on `(tenant_id, legacy_id)` (`20260505101226_AddBasalInjections.cs:52-57`). So on the *second* sync of the same window, re-inserting the same `LegacyId` throws a unique-violation `DbUpdateException`. Because `PublishBasalInjectionsAsync` loops `foreach … await CreateAsync` under a single `catch (Exception)`, the first already-existing injection aborts the whole batch → returns `false` → the sync reports a BasalInjections failure, and any genuinely-new injections later in the list never get persisted.

**Fix:** make `BasalInjection` cover the connector channel the same way `Bolus` already does — give the repository a `LegacyId`-based dedup (e.g. a `BulkCreateAsync` mirroring `BolusRepository`), rather than setting `SyncIdentifier` on connector records (connectors use `LegacyId` by convention; no connector sets `SyncIdentifier`).

### 🟡 Medium — Concentration mis-detection in `ResolveInsulinContext`

The first match attempt is:

```csharp
var match = InsulinCatalog.GetAll().FirstOrDefault(f =>
    normalized.Contains(f.Id, StringComparison.OrdinalIgnoreCase));
```

The catalog lists base IDs before their concentration variants (`humalog` before `humalog-u200`, etc.). A Glooko name like `"Humalog®U200"` normalizes to `"HumalogU200"`, which `.Contains("humalog")` → matches the **U100** entry and never reaches `humalog-u200`. Same for Lyumjev U200, Tresiba U200, Humulin R U500, etc. Result: the wrong `Concentration` (and sometimes `Dia`/`Peak`) is captured into the context, which feeds IOB math. Worth matching most-specific-first, or checking the `U###` suffix before the base name.

### 🟡 Medium — No tests for the new logic

The diff only updates the `TreatmentPublisherTests` constructor to compile. There's no coverage for `MapV3ManualInsulin`, the fuzzy `ResolveInsulinContext`, or `ResolveOrCreatePatientInsulinAsync` — exactly the paths with the subtleties above (and the re-sync behaviour).

### 🟢 Minor

- **Dead code** in `GlookoConnectorService`: `v3Boluses.AddRange(manualBoluses)` runs *after* `v3Boluses` was already published and `v3Boluses` isn't used again, so it has no effect. (Counts stay correct because `PublishRecordTypeAsync` accumulates: `prev + records.Count`.) Suggest removing it to avoid implying it dedupes.
- **PatientInsulin cache keyed by name only, ignoring `Role`** (`BuildPatientInsulinCacheAsync`). A basal injection can resolve to a *bolus*-role `PatientInsulin` (or vice-versa) when names collide, and `IsPrimary = !cache.Values.Any(i => i.Role == role && i.IsPrimary)` ignores existing `Role.Both` primaries, so it can create a second primary — bypassing `SetPrimaryAsync`'s "clear previous primary" logic.
- **Unknown/unmatched insulin persists `PatientInsulinId = Guid.Empty`.** When the name is empty or `"Unknown"`, `ResolveOrCreatePatientInsulinAsync` returns the context unchanged. It's stored as JSON (no FK, so no crash), but downstream IOB resolution by id will silently find nothing. Consider linking a fallback insulin or documenting the intent.
- **Per-record `CreateAsync` in a loop** for basal injections (N round-trips, non-atomic, partial-failure leaves partial state) diverges from the `BulkCreateAsync` pattern every other type uses. Fixing the High finding with a bulk method addresses this too.
- **Enum insertion** — `BasalInjections` added mid-`SyncDataType`. Safe because it's `[JsonStringEnumConverter]` (serialized by name), so no persisted-int drift; placement between `StateSpans`/`Profiles` is just slightly odd vs. grouping near `Boluses`/`CarbIntake`.

---

The basal-injection re-sync bug will surface on the very next import cycle (failing syncs + dropped data), so I'd fix that before merge; the concentration matching and tests are the next priorities.

awilner:
I believe I addressed all the concerns listed above.

PR #367, opened 2026-06-07
feat(api): headless first-tenant setup + migrate instance-key hash to SHA-256

ryceg:
## Summary

Two related changes to the instance-key path:

1. **Headless first-tenant setup.** Trusted instance-key requests (`X-Instance-Key` + `X-Instance-Service`) now bypass the `TenantSetupMiddleware` 503 (`setup_required`/`recovery_mode`) gate. This lets automation stand up a tenant end-to-end — create it, configure connectors, change settings, push data — **before** any human registers a passkey. The human then completes the normal setup wizard exactly as on a fresh instance (the bypass never creates an owner subject, so the wizard's `owner_already_exists` guard still passes). A bare key without the `X-Instance-Service` marker does **not** bypass, mirroring `InstanceKeyHandler`.

2. **Instance-key wire hash SHA-1 → SHA-256.** Hygiene: removes SHA-1 from the one place we control where it isn't externally forced.

## Implementation

- New shared `InstanceKeyValidator` (classifies a request's instance-key credential), used by **both** `InstanceKeyHandler` (auth) and `TenantSetupMiddleware` (gate bypass), so the validation rules — including the service-marker requirement — live in one place.
- SHA-256 applied to every receiver (`InstanceKeyValidator`, `DataHub`) and every sender (web SSR, bridge ×4, demo service). The nocturne-cloud **billing** sender flips in a paired PR on that repo.
- SHA-1 **kept** where an external protocol mandates it — the Nightscout `api-secret` hash and TOTP (`HMAC-SHA1`) — both now annotated as such. Removed the unused `GenerateEventId`.
- README documents the browser and headless setup flows.

## Deploy / coordination

**Hard cutover** — receivers and senders must flip together. The nocturne-cloud billing sender is a separate, coupled PR; deploy both together. For nocturne-cloud, pin `nocturne-api` until the billing image is built, then a single redeploy flips both within the restart window (only billing→api provisioning is exposed, and it's retryable). Self-hosters on the compose bundle get all instance-key images from the same `:latest`, so Watchtower pulls them as one set. No persisted state is keyed on the hash and `INSTANCE_KEY` itself is unchanged, so there is no permanent break and no secret rotation.

## Testing

`dotnet test` (unit): instance-key, setup-middleware (incl. new bypass tests), auth-handler-priority, legacy-grant, api-key suites — all green. Builds clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,35 @@ docker compose up -d
 
 The production compose includes [Watchtower](https://github.com/nicholas-fedor/watchtower) for automatic container updates (checks daily), and omits the Aspire dashboard and Scalar API explorer. Watchtower will automatically pull new images as they are published — no manual updates needed.
 
+### First-run setup
+
+A fresh install has no tenants, so the API answers every request with `503 {"error":"setup_required"}` until the first owner secures the instance. This is expected — it's the signal for the web UI to show the setup wizard.
+
+**In a browser (normal path).** Open your Nocturne site (the apex `https://<BASE_DOMAIN>/`). The setup wizard walks you through creating the first tenant and registering a passkey (or linking an OIDC provider). When it completes you're shown a set of recovery codes — save them. That's the whole setup.
+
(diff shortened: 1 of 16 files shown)


ryceg:
Paired sender change (must deploy together): ryceg/nocturne-cloud#27

PR #410, opened 2026-06-14
fix: current glucose status text type error

DanielDanielsson:
Hello, while setting up nocturne locally I found this type error in `CurrentBGDisplay`. Im using Dexcom share, and a Dexcom G7 sensor (t1d myself) 

CleanShot 2026-06-14 at 11 41 15

## Summary

Fixes the current glucose status text so Svelte receives a concrete string instead of a derived runtime sentinel value.

Adds a shared `minutesAgo()` formatter that uses `Intl.RelativeTimeFormat` with the current preferred language instead of hardcoded English minute text.

Adds a component regression test that renders `CurrentBGDisplay` with a Symbol-valued `timeSinceReading` store field, proving the component no longer reads that value into rendered status text.

## Root cause

`CurrentBGDisplay` wrapped `realtimeStore.timeSinceReading` in another derived value and then passed that value into `GlucoseValueIndicator`. In some runtime paths this produced a Symbol value, which Svelte could not render as text.

## Validation

Passed: `pnpm --dir src/Web/packages/app exec vitest run src/lib/utils/formatting.test.ts`

Checked: `pnpm --dir src/Web/packages/app exec svelte-check-native --tsconfig ./tsconfig.json 2>&1 | rg "CurrentBGDisplay|formatting"` returned no matching diagnostics.

Could not run locally: `pnpm --dir src/Web/packages/app exec vitest --config vitest.browser.config.ts run src/lib/components/dashboard/CurrentBGDisplay.svelte.test.ts` because local Playwright Chromium is not installed.

### DIFF
--- a/src/Web/packages/app/src/lib/components/dashboard/CurrentBGDisplay.svelte
+++ b/src/Web/packages/app/src/lib/components/dashboard/CurrentBGDisplay.svelte
@@ -19,6 +19,7 @@
   import {
     formatGlucoseValue,
     formatGlucoseDelta,
+    minutesAgo,
   } from "$lib/utils/formatting";
   import { Clock } from "lucide-svelte";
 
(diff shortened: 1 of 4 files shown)


DanielDanielsson:
@ryceg I found this bug while testing the Dexcom connection, let me know if this is valid or if I should change anything.

PR #455, opened 2026-07-04
Port tandem source API changes to Nocturne's connector

jwoglom:
**This hasn't been tested** -- but this PR includes a claude fable-generated port of the tconnectsync v3 changes to Nocturne's Tandem Source connector, including all of the same integration tests from the tconnectsync codebase.

Ideally before merging, someone with an active tandem source account containing data in it (I don't have any recent data in my account since i have been testing mobi with trio) should build Nocturne and validate the connector appears to work as expected.


pclarici:
**Tested `pr-455-107ccc6` on a self-hosted instance — the connector works. ✅**

*Transparency note: this test was run end-to-end by Claude (Fable 5) operating my self-hosted instance, and this comment was written by it as well — fitting, since the PR itself is a Fable-generated port.*

Setup: self-hosted docker-compose bundle (v0.1.5 base), Tandem Source **EU** account, connector configured via the web UI.

**Before (release `:latest`):** authentication succeeded but every sync failed in `GetPumpEventMetadataAsync`:

```
System.Text.Json.JsonException: The JSON value could not be converted to System.String.
Path: $[0].tconnectDeviceId — Cannot get the value of a token type 'Number' as a string.
```

**With the PR images** (`nocturne-api` + `nocturne-web` at `pr-455-107ccc6`): migrations applied cleanly and a full historical sync completed with no errors:

```
itemsSynced: Profiles 1, Boluses 1062, CarbIntake 536, BolusCalculations 1062,
TempBasals 52297, DeviceEvents 845, StateSpans 259
```

Also verified: `SyncGlucose=false` is respected (re-sync imported no CGM entries — glucose comes from the Dexcom connector on this instance), manual sync and `reset-cursor` both work, Control-IQ temp basals land correctly, and the counts were verified directly in Postgres, not just from the sync response. Scheduled 5-minute syncs have been picking up new boluses in near-real-time since.

One observation (core behavior, not introduced by this PR): after `DELETE /api/v4/services/connectors/{id}/data` — which soft-deletes — a subsequent re-sync recognizes the tombstoned records as already present and skips them without restoring, so the re-imported history stays invisible until the rows are un-deleted manually. Maintainers may want the admin delete endpoint to hard-delete connector data, or the connector upsert to resurrect tombstones it matches.

ryceg:
Glad to know that fables gonna put us all out of jobs! Great stuff.

jwoglom:
@pclarici thank you for testing!!

PR #465, opened 2026-07-04
feat(identity): cross-tenant caregiver overview endpoint

ryceg:
## Summary

First slice of the multi-tenant caregiver dashboard (a parent monitoring multiple family members' tenants from one place).

Adds `GET /api/v4/me/tenants/overview`: for each tenant the authenticated subject is an active member of **with glucose read permission**, returns:

- latest canonical glucose reading (mg/dL, delta, direction, trend rate, timestamp)
- per-tenant thresholds and a server-side `GlucoseStatus` classification (`Unknown/Stale/UrgentLow/Low/InRange/High/UrgentHigh`)
- staleness (`Overview:StaleAfterMinutes`, default 25)
- active alert excursion count + most severe rule severity (only when the member has `alerts.read`; `null` otherwise)

## Design notes

- **Thresholds are per-tenant**: derived from the tenant's enabled `Threshold` alert rules (Critical → urgent bounds, Warning/Info → low/high, most-conservative merge, ordered clamps so the band can't invert), falling back to global `Thresholds:*` config where no rule exists. Classification lives server-side so every client shows the same status the alert engine would act on.
- **AuthZ mirrors `MemberScopeMiddleware`**: effective permissions = role ∪ direct per membership, intersected with the auth token's granted scopes (superuser `*` membership bypasses the intersection, same as the middleware). A narrowly-scoped token cannot use this endpoint to enumerate BG it couldn't read directly.
- **Tenant isolation**: fan-out never touches the request-scoped DbContext. Canonical glucose reads run in a fresh DI scope per tenant with `ITenantAccessor` pinned (AlertSweepService pattern); rule/excursion queries use a factory context with `TenantId` set before the first query (RLS GUC + EF filter). One tenant failing degrades to an `Unknown` item instead of failing the response.
- **Apex-reachable**: the path is added to `TenantlessAllowedPaths` so the future dashboard can call it from the base domain in multi-tenant deployments.

## Tests

29 new unit tests: threshold resolution (defaults, overrides, most-conservative merge, inverted-band clamp), classification boundaries + staleness, membership filtering (revoked/inactive/no-permission excluded; `*` and readwrite included), token scope gating, alerts-read gating, cross-tenant isolation, per-tenant failure isolation, malformed condition params.

Full unit suite: 3824 passed; 2 failures are pre-existing/flaky and unrelated (`CacheIntegrationTests.CreateEntriesAsync_DecomposesToV4AndFiresEvents` fails on clean main; `StatusServiceTests.GetLastModifiedAsync_ShouldReturnTimestampsForAllCollections` passes in isolation).

No migrations, no generated-client changes.
> Note: an earlier revision added a `caregiver` seed role; it was dropped (strict subset of the existing `caretaker` role) — remote monitors use `caretaker`.

### DIFF
--- a/src/API/Nocturne.API/Controllers/V4/Identity/MyTenantsController.cs
+++ b/src/API/Nocturne.API/Controllers/V4/Identity/MyTenantsController.cs
@@ -4,6 +4,8 @@ using Microsoft.Extensions.Options;
 using OpenApi.Remote.Attributes;
 using Nocturne.API.Authorization;
 using Nocturne.API.Configuration;
+using Nocturne.API.Extensions;
+using Nocturne.API.Services.Identity;
 using Nocturne.Core.Contracts.Multitenancy;
 using Nocturne.Core.Models.Authorization;
 
@@ -25,13 +27,16 @@ namespace Nocturne.API.Controllers.V4.Identity;
 public class MyTenantsController : ControllerBase
 {
     private readonly ITenantService _tenantService;
+    private readonly ITenantOverviewService _overviewService;
     private readonly OperatorConfiguration _config;
 
     public MyTenantsController(
         ITenantService tenantService,
+        ITenantOverviewService overviewService,
         IOptions<OperatorConfiguration> config)
     {
         _tenantService = tenantService;
+        _overviewService = overviewService;
         _config = config.Value;
     }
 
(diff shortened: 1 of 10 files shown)


ryceg:
Dropped the `caregiver` seed role after review feedback: it was a strict subset of the existing `caretaker` role (caretaker already grants glucose read, alerts read/acknowledge, and the device atoms), and the two names are near-synonyms. Remote monitors get `caretaker`. Back to the original six seed roles — no widening was needed.

PR #470, opened 2026-07-05
Re-enable wuchale (0.25) + translation contribution PR endpoint

ryceg:
## Summary

Two pieces of the translation-contribution initiative: the wuchale runtime comes back to life (updated 0.19 → 0.25.6), and the backend gains an endpoint that turns in-app translation contributions into upstream pull requests.

### 1. Re-enable wuchale and update to 0.25

Re-enables the i18n runtime (disabled in b87a8cff over .po merge conflicts and no active translators) and updates it. With wuchale off, the app ships English-only regardless of locale settings.

- **wuchale 0.19.4 → 0.25.6, @wuchale/svelte 0.18.1 → 0.20.4.** The standalone `@wuchale/vite-plugin` is removed — its latest release pins `wuchale ^0.21`, which loads a second copy of wuchale next to 0.25 and crashes the build. The plugin now ships in core (`wuchale/vite`).
- **Config migration** to the 0.25 API: shared `pofile()` storage + top-level `localesDir`. Both configs (app + portal) now list **both** packages' Svelte files, so a single extraction run produces the complete shared catalog. Previously each config's svelte adapter only covered its own package, so extraction runs ping-ponged the catalog (running one obsoleted the other package's messages).
- **All `WUCHALE-DISABLED` sites restored**; loader API rename applied (`loadIDs` → `loadCount`); `loadLocales` awaited so nothing renders before catalogs register (an unloaded runtime silently renders every message as `''`).
- **Portal prerender fix**: the portal is fully static, so a build-time-only `hooks.server.ts` loads catalogs and pins the source locale.
- **Infinite-loop fix**: 0.25's heuristic extracts top-level string constants, which turned `PasswordGenerator`'s `ALPHABET` into a message. Its catalog entry compiled to an empty string, making `ALPHABET.length` 0 → `256 % 0 = NaN` → the rejection-sampling loop spun forever and the portal build never finished. Now `@wc-ignore`d; the catalog-completeness fixes remove the underlying hazard.
- **CI**: `sync-translations.yml` updated for the new CLI (`wuchale sync` → `wuchale`).
- **Catalogs regenerated**: 0.25 message model, ~4.7k messages including portal strings. The old catalogs held only ~50 real translations across all 11 locales, mostly for msgid shapes that no longer exist — nothing meaningful is lost; the Gemini flow can refill.

### 2. Translation contribution endpoint (bot-PR relay)

`POST /api/v4/translations/contributions` accepts a contribution (locale, msgid → msgstr entries, contributor identity, optional note) and opens an upstream PR, following the `GitHubIssueService` pattern:

- **Direct mode** (`GitHub:TranslationsPat` configured): fetch the locale catalog from this repo, apply the entries, commit to a `translations/{locale}-*` branch with a `Co-authored-by` trailer, open a PR listing the contributor, counts, and any entries that no longer match the catalog.
- **Relay mode** (no PAT — the default for self-hosted instances): forward the JSON payload to nocturne.run.
- The relay target is a dedicated **anonymous ingress** (`POST /api/v4/translations/relay`) that 404s unless `GitHub:AcceptRelayedContributions` is enabled and a PAT is present — only the relay host exposes it, and it re-validates and re-rate-limits everything. (Reviewers: the existing support-issue relay posts to an `[Authorize]` endpoint, which would 401 for relayed callers; that flaw is not replicated here and could be fixed the same way in a follow-up.)
- `PoCatalogEditor` rewrites **only msgstr lines of existing entries** (plural-aware, clears fuzzy, header/obsolete untouchable) and preserves the rest of the catalog byte-for-byte — a contribution can never add, remove, or reorder messages. 18 unit tests cover matching, contexts, plurals, escaping, fuzzy clearing, and CRLF preservation.
- Rate-limited (10/IP/hour), `[Authorize]`d, input-bounded (≤500 entries, length caps).

The in-app editor UI that feeds this endpoint is the next PR.

## Verification

- `pnpm build` green for app (adapter-node) and portal (adapter-static, all pages prerendered — previously wedged indefinitely).
- Dev-server SSR of every portal route returns 200.
- `dotnet build` clean; new unit tests 18/18 green.

Re the original disable motivation (.po merge conflicts): CI remains the single writer of catalogs via sync-translations.yml; contribution PRs apply deltas to the current base catalog at submission time rather than merging stale files.


Uneo7:
Tested and valided with a couple of fixes :)

PR #477, opened 2026-07-07
fix(web): make the basal-analysis report show data

Lamercho:
## Problem

The `/reports/basal-analysis` page never shows data, for a chain of reasons:

1. **The page is a stub.** `basalStats`, `tempBasalInfo` and `hourlyPercentiles` are hard-coded to zeros/empty with a comment saying "no dedicated endpoint available" — but `GET /api/v4/statistics/basal-analysis` exists and is fully implemented in `StatisticsController`, and the generated client already ships `getBasalAnalysis`. The stat cards and the percentile chart are therefore always empty.

2. **On instances with realistic data volumes the page renders completely blank** (dark content area, no error, no console output). `getBasalReportData` returns ~1.3 MB here, and consuming a query created inside `$derived` through `.loading`/`.current` (the `contextResource` pattern) hits sveltejs/kit#14915: the resolved response lands on a superseded query instance, `.loading` sticks at `true` forever, and the `{#if reportsResource.current}` guard never opens. Smaller payloads (e.g. the AGP report) usually win the race, which is why only this report appears broken. `insulin-delivery` uses the same query and can hit the same wall.

3. **With data finally flowing, both charts still fail to render.** layerchart `2.0.0-next.43`'s motion tracking is broken on svelte ≥ 5.35.4 (sveltejs/svelte#16548, techniq/layerchart#630): combined with the built-in `legend` prop the page dies with `effect_update_depth_exceeded`; without it the area paths silently stay empty. The AGP chart renders fine because it never passes `motion`.

4. **The percentile chart's median is invisible on every theme**: it uses `hsl(var(--primary))`, but `--primary` is an oklch value, so the color is invalid.

## Changes

- Wire the page to `getBasalAnalysis` and derive `basalStats` / `tempBasalInfo` / `hourlyPercentiles` from the response.
- Move the report body into `BasalAnalysisContent.svelte`, which creates its two query instances once at component init and is remounted via `{#key}` when the date range changes — sidestepping kit#14915 while keeping the `.loading`/`.current` consumption style used elsewhere.
- Drop the `motion` tween props on both report charts (matching the AGP chart, which already renders without motion) and the redundant built-in `legend` on the single-series delivery chart — a workaround until layerchart is bumped to a 2.0.x release containing the motion fix, at which point the tweens can come back.
- `hsl(var(--primary))` → `var(--primary)` for the median line and legend swatch.

## Testing

Built the web image from this branch and ran it against a production instance (API 0.1.6, Nightscout connector, ~16k temp basals / ~5k boluses from a Trio loop):

- stat cards, temp-basal metrics, percentile bands + median line, hourly delivery chart and period insights all render with real data;
- date-range changes remount cleanly;
- no console errors, no server-side errors;
- cross-checked the rendered numbers against raw `GET /api/v4/statistics/basal-analysis` output.

Happy to split this into separate PRs if that's easier to review.

### DIFF
--- a/src/Web/packages/app/src/lib/components/reports/BasalRatePercentileChart.svelte
+++ b/src/Web/packages/app/src/lib/components/reports/BasalRatePercentileChart.svelte
@@ -123,9 +123,7 @@
         seriesLayout="overlap"
         tooltip={{ mode: "bisect-x" }}
         props={{
-          area: { motion: { type: "tween", duration: 200 } },
           xAxis: {
-            motion: { type: "tween", duration: 200 },
             tickMultiline: true,
             format: formatHour,
           },
--- a/src/Web/packages/app/src/lib/components/reports/BasalRatePercentileChart.svelte
+++ b/src/Web/packages/app/src/lib/components/reports/BasalRatePercentileChart.svelte
@@ -47,7 +47,7 @@
     .range([
       "var(--chart-1)",
       "var(--chart-2)",
-      "hsl(var(--primary))",
+      "var(--primary)",
     ]);
 </script>
 
@@ -93,7 +93,7 @@
               (d: HourlyBasalPercentileData) => d.median,
               (d: HourlyBasalPercentileData) => d.median,
             ],
-            color: "hsl(var(--primary))",
+            color: "var(--primary)",
             props: {
               line: { strokeWidth: 2 },
             },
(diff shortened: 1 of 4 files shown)


ryceg:
@Lamercho thanks for this- but you need to sign the CLA in order for it to be mergeable :)

Lamercho:
But I did

PR #525, opened 2026-07-19
fix(v1/v3): present record ids as 24-hex Mongo ObjectIds for AAPS (#522)

### LINKED ISSUE #522: NumberFormatException During AndroidAPS Sync
Image

#### Summary

AndroidAPS (AAPS) crashed with `java.lang.NumberFormatException` during downward syncs from Nocturne. The failure occurred when AAPS attempted to parse a non-numeric record ID as a `Long`.

How to Reproduce:
Connect AndroidAPS and upload a new treatment (e.g., a carb entry or bolus). Nocturne’s backend will ingest the treatment and pass it through and save it to the database. On the very next synchronization cycle, AAPS will fetch this new record, fail the strict 24-character isObjectId() validation check, crashing the sync loop.

ryceg:
Closes #522.

AAPS validates every record id with a strict `isObjectId()` check (`[0-9a-f]{24}`) and, on failure, parses the id as a `Long`. A Nocturne UUID (36 chars) satisfies neither, so **every AAPS-uploaded treatment/entry crashed the next downward sync** with `NumberFormatException`.

## Emit side — 24-hex ObjectId on the wire
`MongoObjectId.Coerce(id)` maps a UUID to the first 24 hex of its canonical form (a real ObjectId passes through; any other string hashes deterministically to 24 hex). Applied via a `JsonConverter` on `Treatment._id`/`identifier`, the `Entry` V1/V3 response wrappers, and inline in the V3 devicestatus dto, plus the V3 create/dedup/patch `identifier` fields and the `Location` route. **Only the wire representation changes — the in-memory value is untouched**, so internal logic and the `LegacyId` dedup path are unaffected (`Read` is a passthrough).

## Reverse resolution — full round-trip
A 24-hex ObjectId derived from a UUID resolves back via a uuid **prefix range** `[oid+"00000000", oid+"ffffffff"]`. Postgres orders `uuid` byte-wise (= hex-prefix order), so the range selects exactly the source record. New `GetByGuidRangeAsync` on `IV4Repository`/`V4RepositoryBase` (+ the four non-base repos: TempBasal, Aps/Pump/Uploader snapshot, BasalInjection), wired into `GetByIdAsync` and `DELETE` for treatments, entries, and devicestatus, and into dedup-on-create.

## Update in place — no duplicates
`DecomposeAsync` upserts by `LegacyId`, but the wire id is a derived ObjectId, so a naive re-decompose would insert a duplicate. `ResolveCanonicalIdAsync` maps the wire ObjectId back to the stored `LegacyId`; `UpdateAsync`/`PatchTreatmentAsync` re-key to it before re-decomposing, and `ApplyJsonPatch` now preserves the id across its serialize round-trip (a PATCH never changes identity). Temp basals emit their UUID when the `LegacyId` isn't already an ObjectId, so temp-basal PATCH resolves via range.

## Why a range query (and why it's tested against real Postgres)
The reverse mapping is lossy (24 hex from a 128-bit UUID), so it can't be inverted arithmetically — it relies on Postgres `uuid` ordering matching hex-prefix ordering. **.NET's own `Guid` ordering differs**, so the InMemory provider can't validate it; the equivalence is pinned by an integration test against a real Postgres container (V4 goldens).

## Tests
- `MongoObjectIdTests` — helper + `Treatment`/`Entry` serialization (unit).
- `ObjectIdRangeResolutionGoldenTests` — uuid-prefix-range resolution on **real Postgres** (integration).
- Full unit suite green (4070 API + 332 Core.Models + 505 Infra), no regressions.

## Known limitation (pre-existing, tracked separately)
A native, UI-authored treatment that never received a `LegacyId` still can't be updated in place by AAPS (it duplicates) — the decomposer has no key to match. This predates the fix (all treatment PATCHes duplicated before, since the projected id was the UUID, not the `LegacyId`); the re-key here fixes the common case where AAPS edits its own (`LegacyId`-bearing) records. Fully closing it needs the decomposer to accept a UUID/ObjectId match key.

Credit to @old-square-eyes for the diagnosis in #522.

### DIFF
--- a/src/API/Nocturne.API/Controllers/V3/DeviceStatusController.cs
+++ b/src/API/Nocturne.API/Controllers/V3/DeviceStatusController.cs
@@ -700,7 +700,8 @@ public class DeviceStatusController : BaseV3Controller<DeviceStatus>
             ["created_at"] = status.CreatedAt,
             ["uploaderBattery"] = status.Uploader?.Battery,
             ["utcOffset"] = status.UtcOffset,
-            ["identifier"] = status.Id,
+            // 24-hex ObjectId so AAPS's isObjectId() passes; resolved back via GetByIdAsync.
+            ["identifier"] = MongoObjectId.Coerce(status.Id),
             ["srvModified"] = status.Mills,
             ["srvCreated"] = status.Mills,
         };
--- a/src/API/Nocturne.API/Controllers/V3/EntriesController.cs
+++ b/src/API/Nocturne.API/Controllers/V3/EntriesController.cs
@@ -234,13 +234,14 @@ public class EntriesController : BaseV3Controller<Entry>
                 );
                 if (existingEntry != null)
                 {
+                    var existingIdentifier = MongoObjectId.Coerce(existingEntry.Id);
                     return Ok(
                         new
                         {
                             status = 200,
-                            identifier = existingEntry.Id,
+                            identifier = existingIdentifier,
                             isDeduplication = true,
-                            deduplicatedIdentifier = existingEntry.Id,
+                            deduplicatedIdentifier = existingIdentifier,
                         }
                     );
                 }
@@ -267,12 +268,10 @@ public class EntriesController : BaseV3Controller<Entry>
 
             await EvaluateAlertsAsync(new[] { createdEntry }, cancellationToken);
 
-            // Set location header for created resource
-            Response.Headers["Location"] = $"/api/v3/entries/{createdEntry.Id}";
-
+            // Location resolves to the 24-hex ObjectId that matches the response body identifier.
             return CreatedAtAction(
                 nameof(GetEntry),
-                new { id = createdEntry.Id },
+                new { id = MongoObjectId.Coerce(createdEntry.Id) },
                 createdEntry.ToV3Response()
             );
         }
--- a/src/API/Nocturne.API/Controllers/V3/TreatmentsController.cs
+++ b/src/API/Nocturne.API/Controllers/V3/TreatmentsController.cs
@@ -222,13 +222,14 @@ public class TreatmentsController : BaseV3Controller<Treatment>
                 );
                 if (existingTreatment != null)
                 {
+                    var existingIdentifier = MongoObjectId.Coerce(existingTreatment.Id);
                     return Ok(
                         new
                         {
                             status = 200,
-                            identifier = existingTreatment.Id,
+                            identifier = existingIdentifier,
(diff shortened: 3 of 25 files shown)


ryceg:
## Adversarial review pass

Ran three adversarial reviewers (duplicates/data-loss, multi-tenancy/security, serialization blast-radius). Two real data-integrity bugs found and fixed (commit `48a6a00`); everything else cleared or noted.

**Fixed (HIGH):**
- **Meal-bolus DELETE orphaned the carb.** Deleting by the derived wire ObjectId used the single-row range delete, leaving the correlated carb behind as a phantom correction. Now resolves to the record's `LegacyId` and deletes via `DeleteByLegacyId` (removes all correlated siblings).
- **PATCH of a native `LegacyId==null` row duplicated it.** The decomposer upserts by `LegacyId`; a null one couldn't match → guaranteed duplicate for Note/BGCheck/DeviceEvent/BolusCalc. `ResolveCanonicalIdAsync` now backfills the null `LegacyId` with the derived ObjectId so the update lands in place.

**Cleared with evidence:**
- **Security/tenancy** — no cross-tenant leak, RLS bypass, soft-delete bypass, or scope bypass: `GetByGuidRangeAsync` reuses the same tenant-pinned, soft-delete-filtered `ITenantDbContextFactory` with no `IgnoreQueryFilters`, backed by DB RLS. No unguarded parse/exception path.
- **Serialization blast radius** — cache (stores object refs, not JSON), connector republish, OpenAPI/frontend (frontend uses V4 shapes), synthetic-id/dedup, SignalR, MCP, analytics all cleared. The one dangerous internal round-trip (`ApplyJsonPatch`) was already guarded.
- Probe-order collision, hash collision, and the `existing`/re-key asymmetry are all ~2⁻⁹⁶ under UUIDv7 — unreachable. Added `OrderBy(e => e.Id)` for determinism anyway.

**Noted, not blocking (follow-ups):**
- *Authenticated query amplification (MEDIUM):* a non-UUID lookup fans out to ~14 sequential PK-indexed queries (7 legacy-probe + 7 range-probe); reachable on POST-dedup with a crafted `_id`. Index-backed and bulk-exempt, so it's request-amplification, not a scan DoS. Could be reduced later by resolving the type once.
- *DDataService clones (LOW/latent):* deep-clone-via-serialize now yields the ObjectId id in-memory; harmless today (read-only pipeline, re-coerced on the wire) and largely defused by the reverse-resolution.

PR #578, opened 2026-07-26
fix(auth): authenticate the OIDC state parameter

ryceg:
## The bug

The `state` parameter is plain base64url JSON. No signature, no encryption, no server-side record:

```csharp
private static string EncodeState(OidcStateData data)
{
    var json = JsonSerializer.Serialize(data);
    var bytes = Encoding.UTF8.GetBytes(json);
    return Base64Url.Encode(bytes);
}
```

The only integrity check is `state != expectedState`, where `expectedState` comes from the `.Nocturne.OidcState` cookie. That's a double-submit CSRF defence — it stops a *victim's browser* being walked through a flow it didn't start. It does nothing against a caller acting as its own HTTP client, which sets both halves. So every field in the state is caller-chosen.

The callback trusts two of them. `Intent` selects which flow runs; `SubjectId` names the account an external identity gets bound to. `HandleSetupCallbackAsync` reads both:

```csharp
if (stateData.Intent != "setup") return Failed("invalid_intent", ...);
if (!stateData.SubjectId.HasValue) return Failed("invalid_state", ...);
var subjectId = stateData.SubjectId.Value;

var (outcome, _) = await _subjectService.AttachOidcIdentityAsync(subjectId, provider.Id, claims.Sub, ...);
```

Given `{"Intent":"setup","SubjectId":"<any guid>","Nonce":null}`, that links the caller's own provider account to the named subject and issues a session for it. `AttachOidcIdentityAsync` returns `Created` for a fresh provider account — only `AlreadyLinkedToOther` fails — so the bind succeeds against a subject the caller has never had access to. Setting `Nonce` to null also skipped the ID-token nonce check, which was conditional on the state carrying one.

The endpoint it lands on, `/api/v4/setup/oidc/callback`, is `[AllowAnonymous] [AllowDuringSetup]` and had no setup-state gate, so it stays reachable for the life of the instance. Naming the platform admin's subject yields a session as the platform admin, and `PlatformAccessController` then mints `permissions: ["*"]` into any tenant by slug — cross-tenant CGM and therapy data.

Preconditions are just: an enabled provider whose registered redirect URIs include the setup callback (true of any instance whose owner set up via OIDC), and knowledge of a target subject GUID.

## The fix

Protect the payload with `IDataProtector`:

```csharp
private string EncodeState(OidcStateData data) =>
    _stateProtector.Protect(JsonSerializer.Serialize(data));
```

This is the pattern `PasskeyService`, `TotpService`, `OidcProviderService` and `GuestSessionHandler` already use for client-held blobs, and keys are already persisted to the database (`AddDataProtection().PersistKeysToNocturneDb()`), so it survives restarts and works across instances. `EncodeState`/`DecodeState` are the single chokepoint, so login, link and setup are all covered by the one change — no per-flow patching.

Two supporting changes:

- **The nonce check is now unconditional** for OIDC providers. Every state this service issues carries a nonce, so an absent one means the ID token can't be bound to the authorization request. No behaviour change for a compliant provider — one that didn't echo the nonce already failed the comparison.
- **Dropped the caller-supplied `state` override** on `GenerateAuthorizationUrlAsync`. No caller passed it, and it was a route to putting an unprotected state back on the wire.

### Why not a server-side state store

It's the other standard answer and it's strictly more moving parts: a table, a write on every authorization start, and expiry pruning. Protecting the payload gets the same property — the callback can only act on state this instance issued — with no new storage. If we later want single-use states (replay within the 5-minute lifetime), that's the point to add a store.

## Deploy note

OIDC logins already in flight when this rolls out fail with `invalid_state` and succeed on retry. Nothing persisted is affected; no migration.

## Testing

Full API unit suite green: **4175 passed, 0 failed**.

Five new tests in `OidcAuthServiceStateProtectionTests`:

| Test | Asserts |
|---|---|
| `HandleSetupCallback_WithForgedState_IsRejected` | the account-takeover primitive, closed |
| `HandleCallback_WithForgedState_IsRejected` | login callback |
| `HandleLinkCallback_WithForgedState_IsRejected` | link callback |
| `GeneratedSetupState_DoesNotExposeItsPayload` | the subject isn't legible in the state |
| `StateFromAnotherKeyRing_IsRejected` | a well-formed state this instance didn't issue is refused |

The forged-state helper builds exactly what the old `EncodeState` produced, so the tests encode the real attack rather than a proxy for it. I checked the first four genuinely fail against the plain-encoding implementation (the fifth is the one I added *because* an earlier tamper-a-character test passed either way, which made it worthless as a regression guard).

## Related

nightscout/nocturne#577 adds the matching defence in depth on the setup side: a setup-state gate on `OidcCallback`, and a platform-admin grant that re-derives the first owner instead of trusting the subject it's handed. That PR holds the line if this one ever regresses; this PR is the root cause.

### DIFF
--- a/src/API/Nocturne.API/Middleware/OidcCallbackRedirectMiddleware.cs
+++ b/src/API/Nocturne.API/Middleware/OidcCallbackRedirectMiddleware.cs
@@ -1,8 +1,6 @@
-using System.Text;
-using System.Text.Json;
 using Microsoft.Extensions.Options;
-using Nocturne.API.Helpers;
 using Nocturne.API.Multitenancy;
+using Nocturne.Core.Contracts.Auth;
 
 namespace Nocturne.API.Middleware;
 
@@ -21,8 +19,10 @@ namespace Nocturne.API.Middleware;
 /// <see cref="MemberScopeMiddleware"/>, <see cref="SiteSecurityMiddleware"/>.
 /// </para>
 /// <para>
-/// Extracts the tenant slug from the base64-encoded OIDC <c>state</c> query parameter
-/// and issues a 302 redirect to the correct <c>{slug}.{baseDomain}</c> URL.
+/// Reads the tenant slug out of the protected OIDC <c>state</c> query parameter via
+/// <see cref="IOidcAuthService.TryReadTenantSlug"/> and issues a 302 redirect to the correct
+/// <c>{slug}.{baseDomain}</c> URL. The slug comes from the authenticated state rather than
+/// from a caller-supplied parameter, so the redirect target is one this instance issued.
 /// </para>
 /// </remarks>
 /// <seealso cref="Multitenancy.TenantResolutionMiddleware"/>
@@ -88,7 +88,8 @@ public partial class OidcCallbackRedirectMiddleware
             return;
         }
 
-        var tenantSlug = ExtractTenantSlug(stateParam);
+        var oidcAuthService = context.RequestServices.GetRequiredService<IOidcAuthService>();
+        var tenantSlug = oidcAuthService.TryReadTenantSlug(stateParam);
         if (string.IsNullOrEmpty(tenantSlug))
         {
             await _next(context);
@@ -116,25 +117,6 @@ public partial class OidcCallbackRedirectMiddleware
     {
         var value = path.Value ?? "";
         return CallbackPaths.Any(p => value.Equals(p, StringComparison.OrdinalIgnoreCase));
-    }
-
-    private static string? ExtractTenantSlug(string encoded)
-    {
-        try
-        {
-            var bytes = Base64Url.Decode(encoded);
-            var json = Encoding.UTF8.GetString(bytes);
-
-            using var doc = JsonDocument.Parse(json);
-            if (doc.RootElement.TryGetProperty("TenantSlug", out var prop))
-                return prop.GetString();
-
-            return null;
-        }
-        catch
-        {
-            return null;
-        }
     }
 
     [System.Text.RegularExpressions.GeneratedRegex(@"^[a-z0-9][a-z0-9\-]{0,61}[a-z0-9]$")]
--- a/src/API/Nocturne.API/Services/Auth/OidcAuthService.cs
+++ b/src/API/Nocturne.API/Services/Auth/OidcAuthService.cs
@@ -3,6 +3,7 @@ using System.Security.Cryptography;
 using System.Text;
 using System.Text.Json;
 using System.Text.Json.Serialization;
+using Microsoft.AspNetCore.DataProtection;
 using Microsoft.Extensions.Options;
 using Nocturne.API.Helpers;
 using Nocturne.Core.Constants;
@@ -29,6 +30,7 @@ public class OidcAuthService : IOidcAuthService
     private readonly IRefreshTokenService _refreshTokenService;
     private readonly IHttpClientFactory _httpClientFactory;
     private readonly ITenantMemberService _tenantMemberService;
+    private readonly IDataProtector _stateProtector;
     private readonly OidcOptions _options;
     private readonly IConfiguration _configuration;
     private readonly ILogger<OidcAuthService> _logger;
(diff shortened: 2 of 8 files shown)


ryceg:
## Live verification + security review

### Live test against a running stack

Booted the stack from this branch with a real OIDC provider row pointed at Google's actual discovery document, and exercised the flow end to end.

| Check | Result |
|---|---|
| Authorization URL built from Google's real discovery doc | 302 issued |
| State URL-safe — appears **verbatim** in `Location` while `redirect_uri` beside it *is* percent-encoded | `^[A-Za-z0-9_-]+$` |
| State length / full authorize URL | 475 / 729 chars |
| Cookie value == URL state value | identical |
| Payload decodable as plain JSON | no, opaque |
| Genuine state round-trips and drives a real HTTPS token exchange | `token_exchange_failed` (fake code rejected by Google) |
| Tampered state | `invalid_state` |
| **Forged plain-JSON setup state naming a victim subject**, attacker supplying both cookie and query param | `invalid_state`, **no session cookies issued** |

That closes the two risks flagged as unverified in review: base64url output passes through the redirect untouched, and 729 chars is well inside any real limit.

Also verified the apex→subdomain bounce, which is the seam the second commit repairs: a login started on `livetest.nocturne.localhost` mints a 496-char slug-carrying state, and the apex callback correctly 302s to `http://livetest.nocturne.localhost/api/auth/oidc/callback?code=…&state=…` with both preserved — while a slug-less apex state correctly does not redirect to a subdomain.

### Review finding — one blocking regression, now fixed (48171cd57)

The review caught something the diff hid: commit `3ed346c8f` also rewrote the cross-tenant membership gate in `CompleteLoginAsync` to run only when the state carried a `TenantSlug`. That was not in its commit message and not in this PR description, and it re-opened the hole the guard exists to close.

`TenantSlug` empty + tenant resolved performed no check at all, and that combination is a first-class flow rather than a corner:

- `/api/auth/oidc/login` is in `TenantlessAllowedPaths`, so an apex-initiated login resolves no tenant and mints a state with `TenantSlug = null` — the platform-access operator bounce, by design.
- `GetRedirectUri` builds the redirect URI from the configured `BaseUrl`, not the request host, so the token exchange succeeds regardless of which host receives the callback.
- A callback delivered to `{tenant}.{basedomain}` resolves that tenant, and `OidcCallbackRedirectMiddleware` passes subdomain requests straight through.

So an attacker authenticating with their own legitimate provider account could start at the apex and replay the callback at a victim's subdomain to be issued a session there, unchecked. `AuthenticationMiddleware` still blocks the data per request, so this is a lost backstop rather than direct PHI access — but the guard's own comment is explicit that blocking only after a session and a "logged in" UI exist is not good enough.

I reproduced the precondition live before fixing: an apex-minted state (475 chars, no slug) is accepted at a tenant subdomain and reaches the token exchange, the step immediately before this gate.

Both conditions now stand, since either alone leaves a gap — a resolved tenant always requires membership, and a state naming a tenant requires one to have resolved.

Four tests added, all verified failing against the broken form:

- `CompleteLoginAsync_WhenStateHasNoSlugButATenantResolved_StillChecksMembership`
- `CompleteLoginAsync_WhenStateNamesATenantButNoneResolved_IsDenied`
- `TryReadTenantSlug_RoundTripsTheSlugItWasIssuedWith` — closes the seam the middleware tests stub out, which is how the earlier break went unnoticed
- `TryReadTenantSlug_ForAStateThisInstanceDidNotIssue_ReturnsNull`

Worth recording how the first vacuity check misled me: it reported all tests passing with the regression reinstated, because the running Aspire `dotnet watch` held this worktree's API assembly and the test run used a stale build. Stopping the stack and rebuilding clean produced the expected 2 failures.

### Verified sound by the review

Exhaustive sweep for a second state producer or consumer — the failure mode that bit this PR once — found none: `DecodeState` has exactly two call sites and `EncodeState` one, all three URL generators funnel through `BuildAuthorizationUrlAsync`, controllers treat state as an opaque string, the generated clients and `sdk/openapi-v4.json` type it as a bare `string`, the bot/bridge and billing use entirely separate state schemes, and no consumer exists in nocturne-cloud. Also confirmed: distinct protector purposes, fail-closed on every decode path, nonce present at all three construction sites with OAuth2 correctly excluded, and DB-persisted keys so rotation and multi-instance work.

Full API suite green: **4180 passed, 0 failed**.

ryceg:
Superseded by #580.

Closing rather than merging: this branch accumulated commits I could not vouch for. Commit `3ed346c8f` in particular carried a rewrite of the cross-tenant membership gate in `CompleteLoginAsync` that was not described in its own commit message and was not in the PR description — it was swept in by a `git add -A` without review, and it re-opened the hole that guard exists to close. That is not a history worth landing on `main`, whatever the final tree looks like.

#580 has identical content, re-authored as three commits after reading all 808 lines of the diff against `main` hunk by hunk. I verified the trees match byte for byte before pushing it.

Nothing here is lost. The vulnerability analysis, the live-test results against a real IdP, and the security review's findings are all carried over into #580's description, and the detailed write-up in the comment above remains readable on this PR.

PR #580, opened 2026-07-26
fix(auth): authenticate the OIDC state parameter

ryceg:
Supersedes #578, which carried some commits I could not vouch for. Same content, re-authored in three reviewed commits. The analysis and live-test results on #578 remain valid and are summarised below.

## The bug

The `state` parameter is plain base64url JSON. No signature, no encryption, no server-side record:

```csharp
private static string EncodeState(OidcStateData data)
{
    var json = JsonSerializer.Serialize(data);
    var bytes = Encoding.UTF8.GetBytes(json);
    return Base64Url.Encode(bytes);
}
```

The only integrity check is `state != expectedState`, where `expectedState` comes from the `.Nocturne.OidcState` cookie. That is a double-submit CSRF defence — it stops a *victim's browser* being walked through a flow it did not start. It does nothing against a caller acting as its own HTTP client, which sets both halves. So every field in the state is caller-chosen.

The callback trusts two of them. `Intent` selects which flow runs; `SubjectId` names the account an external identity gets bound to. `HandleSetupCallbackAsync` reads both:

```csharp
if (stateData.Intent != "setup") return Failed("invalid_intent", ...);
if (!stateData.SubjectId.HasValue) return Failed("invalid_state", ...);
var subjectId = stateData.SubjectId.Value;

var (outcome, _) = await _subjectService.AttachOidcIdentityAsync(subjectId, provider.Id, claims.Sub, ...);
```

Given `{"Intent":"setup","SubjectId":"<any guid>","Nonce":null}`, that links the caller's own provider account to the named subject and issues a session for it. `AttachOidcIdentityAsync` returns `Created` for a fresh provider account — only `AlreadyLinkedToOther` fails — so the bind succeeds against a subject the caller has never had access to. Setting `Nonce` to null also skipped the ID-token nonce check, which was conditional on the state carrying one.

It lands on `/api/v4/setup/oidc/callback`, which is `[AllowAnonymous] [AllowDuringSetup]` and stays reachable for the life of the instance. Naming the platform admin's subject yields a session as the platform admin, and `PlatformAccessController` then mints `permissions: ["*"]` into any tenant by slug — cross-tenant CGM and therapy data.

Preconditions: an enabled provider whose registered redirect URIs include the setup callback, and knowledge of a target subject GUID. Subject ids are UUIDv7 and not brute-forceable, but they are not secrets — `POST /api/oauth/introspect` is unauthenticated and returns `sub` for any token.

## The change

**1 — Protect the payload** (`fedc1d893`). `EncodeState`/`DecodeState` now go through `IDataProtector`, matching `PasskeyService`, `TotpService`, `OidcProviderService` and `GuestSessionHandler`. Keys already persist to the database (`AddDataProtection().PersistKeysToNocturneDb()`), so this survives restarts and works across instances. It is a single chokepoint, so login, link and setup are all covered by one change — no per-flow patching. Also makes the nonce check unconditional for OIDC providers, and drops the unused caller-supplied `state` override that was a route to putting an unprotected state back on the wire.

**2 — One decoder** (`5325a6553`). `OidcCallbackRedirectMiddleware` kept its *own* copy of the state decoder to extract `TenantSlug` for the apex-to-subdomain bounce. Protecting the payload leaves that copy reading a blob it cannot parse — it swallows the failure, returns null, and apex callbacks silently stop redirecting to the tenant subdomain, while still compiling. `TryReadTenantSlug` routes it through the same protector. It exposes only the slug, and the existing regex still guards the redirect URL.

**3 — Tighten the cross-tenant guard** (`a21dfa715`). The guard denied non-members of a resolved tenant but did nothing when the state named a tenant and none resolved — the exact signature of the bounce in (2) failing. Both conditions now stand, because either alone leaves a gap: a resolved tenant always requires membership (the redirect URI is built from the configured base URL, not the request host, so an apex-minted state replayed at `{tenant}.{basedomain}` arrives with a tenant resolved), and a state naming a tenant requires one to have resolved.

### Why not a server-side state store

The other standard answer, and strictly more moving parts: a table, a write per authorization start, expiry pruning. Protecting the payload gets the same property — the callback can only act on state this instance issued — with no new storage. If single-use states are wanted later (replay inside the 15-minute lifetime), that is the point to add one.

## Deploy note

OIDC logins already in flight when this rolls out fail with `invalid_state` and succeed on retry. Nothing persisted is affected; no migration.

## Live verification

Run against a real stack with a provider pointed at Google's actual discovery document:

| Check | Result |
|---|---|
| Authorization URL built from Google's real discovery doc | 302 issued |
| State URL-safe — appears **verbatim** in `Location` while `redirect_uri` beside it *is* percent-encoded | `^[A-Za-z0-9_-]+$` |
| State length / full authorize URL | 475 / 729 chars |
| Cookie value == URL state value | identical |
| Payload decodable as plain JSON | no, opaque |
| Genuine state round-trips and drives a real HTTPS token exchange | `token_exchange_failed` (fake code rejected by Google) |
| Tampered state | `invalid_state` |
| **Forged plain-JSON setup state naming a victim subject**, attacker supplying both cookie and query param | `invalid_state`, **no session cookies issued** |

The apex-to-subdomain bounce was verified too: a login started on `livetest.nocturne.localhost` mints a 496-char slug-carrying state, and the apex callback correctly 302s to `http://livetest.nocturne.localhost/api/auth/oidc/callback?code=…&state=…` with both preserved — while a slug-less apex state correctly does not redirect to a subdomain.

## Testing

Full API unit suite: **4180 passed, 0 failed**.

| Test | Asserts |
|---|---|
| `HandleSetupCallback_WithForgedState_IsRejected` | the account-takeover primitive, closed |
| `HandleCallback_WithForgedState_IsRejected` | login callback |
| `HandleLinkCallback_WithForgedState_IsRejected` | link callback |
| `GeneratedSetupState_DoesNotExposeItsPayload` | the subject is not legible in the state |
| `StateFromAnotherKeyRing_IsRejected` | a well-formed state this instance did not issue is refused |
| `TryReadTenantSlug_RoundTripsTheSlugItWasIssuedWith` | closes the seam the middleware tests stub out |
| `TryReadTenantSlug_ForAStateThisInstanceDidNotIssue_ReturnsNull` | forged and malformed input |
| `CompleteLoginAsync_WhenStateHasNoSlugButATenantResolved_StillChecksMembership` | membership keyed on the resolved tenant, not the state |
| `CompleteLoginAsync_WhenStateNamesATenantButNoneResolved_IsDenied` | a failed bounce is unverifiable, not tenantless |

The forged-state helper builds exactly what the old `EncodeState` produced, so the tests encode the real attack rather than a proxy for it. Each new guard was verified by removing it and confirming the corresponding tests fail.

One caveat worth recording: a vacuity check run while an Aspire stack was live against the same worktree reported false passes, because `dotnet watch` held the API assembly and the test run used a stale build. The verifications above were re-run against a clean build.

## Prior review coverage

A security review of this content swept exhaustively for a second state producer or consumer — the failure mode that caused (2) — and found none: `DecodeState` has exactly two call sites and `EncodeState` one, all three URL generators funnel through `BuildAuthorizationUrlAsync`, controllers treat state as an opaque string, the generated clients and `sdk/openapi-v4.json` type it as a bare `string`, the bot/bridge and billing use entirely separate state schemes, and no consumer exists in nocturne-cloud. Also confirmed: distinct protector purposes, fail-closed on every decode path, a nonce at all three construction sites with OAuth2 correctly excluded, and DB-persisted keys so rotation and multi-instance work.

## Related

- #577 — setup-state gate on the OIDC callback, and re-deriving the platform-admin grant
- #579 — anonymous passkey binding to any subject, and an unauthenticated demo admin API

Same class in all three: a bootstrap path that stays live and trusts a caller-supplied identity.

### DIFF
--- a/src/API/Nocturne.API/Middleware/OidcCallbackRedirectMiddleware.cs
+++ b/src/API/Nocturne.API/Middleware/OidcCallbackRedirectMiddleware.cs
@@ -1,8 +1,6 @@
-using System.Text;
-using System.Text.Json;
 using Microsoft.Extensions.Options;
-using Nocturne.API.Helpers;
 using Nocturne.API.Multitenancy;
+using Nocturne.Core.Contracts.Auth;
 
 namespace Nocturne.API.Middleware;
 
@@ -21,8 +19,10 @@ namespace Nocturne.API.Middleware;
 /// <see cref="MemberScopeMiddleware"/>, <see cref="SiteSecurityMiddleware"/>.
 /// </para>
 /// <para>
-/// Extracts the tenant slug from the base64-encoded OIDC <c>state</c> query parameter
-/// and issues a 302 redirect to the correct <c>{slug}.{baseDomain}</c> URL.
+/// Reads the tenant slug out of the protected OIDC <c>state</c> query parameter via
+/// <see cref="IOidcAuthService.TryReadTenantSlug"/> and issues a 302 redirect to the correct
+/// <c>{slug}.{baseDomain}</c> URL. The slug comes from the authenticated state rather than
+/// from a caller-supplied parameter, so the redirect target is one this instance issued.
 /// </para>
 /// </remarks>
 /// <seealso cref="Multitenancy.TenantResolutionMiddleware"/>
@@ -88,7 +88,8 @@ public partial class OidcCallbackRedirectMiddleware
             return;
         }
 
-        var tenantSlug = ExtractTenantSlug(stateParam);
+        var oidcAuthService = context.RequestServices.GetRequiredService<IOidcAuthService>();
+        var tenantSlug = oidcAuthService.TryReadTenantSlug(stateParam);
         if (string.IsNullOrEmpty(tenantSlug))
         {
             await _next(context);
@@ -116,25 +117,6 @@ public partial class OidcCallbackRedirectMiddleware
     {
         var value = path.Value ?? "";
         return CallbackPaths.Any(p => value.Equals(p, StringComparison.OrdinalIgnoreCase));
-    }
-
-    private static string? ExtractTenantSlug(string encoded)
-    {
-        try
-        {
-            var bytes = Base64Url.Decode(encoded);
-            var json = Encoding.UTF8.GetString(bytes);
-
-            using var doc = JsonDocument.Parse(json);
-            if (doc.RootElement.TryGetProperty("TenantSlug", out var prop))
-                return prop.GetString();
-
-            return null;
-        }
-        catch
-        {
-            return null;
-        }
     }
 
     [System.Text.RegularExpressions.GeneratedRegex(@"^[a-z0-9][a-z0-9\-]{0,61}[a-z0-9]$")]
--- a/src/API/Nocturne.API/Services/Auth/OidcAuthService.cs
+++ b/src/API/Nocturne.API/Services/Auth/OidcAuthService.cs
@@ -3,6 +3,7 @@ using System.Security.Cryptography;
 using System.Text;
 using System.Text.Json;
 using System.Text.Json.Serialization;
+using Microsoft.AspNetCore.DataProtection;
 using Microsoft.Extensions.Options;
 using Nocturne.API.Helpers;
 using Nocturne.Core.Constants;
@@ -29,6 +30,7 @@ public class OidcAuthService : IOidcAuthService
     private readonly IRefreshTokenService _refreshTokenService;
     private readonly IHttpClientFactory _httpClientFactory;
     private readonly ITenantMemberService _tenantMemberService;
+    private readonly IDataProtector _stateProtector;
     private readonly OidcOptions _options;
     private readonly IConfiguration _configuration;
     private readonly ILogger<OidcAuthService> _logger;
@@ -43,6 +45,7 @@ public class OidcAuthService : IOidcAuthService
     /// <param name="refreshTokenService">Service for validating refresh tokens (non-rotation refresh path only).</param>
     /// <param name="httpClientFactory">Factory for the <c>OidcProvider</c> named HTTP client.</param>
     /// <param name="tenantMemberService">Service for verifying tenant membership before issuing a login session.</param>
+    /// <param name="dataProtectionProvider">Provider for the protector that authenticates the state parameter.</param>
     /// <param name="options">OIDC session and state configuration options.</param>
     /// <param name="configuration">Application configuration for reading the base URL.</param>
     /// <param name="logger">Logger instance.</param>
@@ -54,6 +57,7 @@ public class OidcAuthService : IOidcAuthService
         IRefreshTokenService refreshTokenService,
         IHttpClientFactory httpClientFactory,
         ITenantMemberService tenantMemberService,
+        IDataProtectionProvider dataProtectionProvider,
         IOptions<OidcOptions> options,
         IConfiguration configuration,
         ILogger<OidcAuthService> logger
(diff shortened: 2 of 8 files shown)


ryceg:
## Review of the cross-tenant guard: **REGRESSION FIXED — SHIP**, no blocking issues

A review was commissioned specifically to check that commit `a21dfa715` closes the gap rather than papering over it, and — separately — that it does not lock anyone out. Both answered.

### The guard, all four combinations

| `TenantSlug` | `currentTenantId` | Outcome |
|---|---|---|
| empty | null | Session issued, no check — **intended**; apex login with nothing to check against, and `AuthenticationMiddleware`'s per-request member gate still applies to any tenant data |
| empty | resolved | `IsMemberAsync` → `NotAMember` before session issuance — **this is the pair that was broken** |
| set | resolved | Same branch, membership enforced |
| set | null | Hard `invalid_state` |

`stateData.TenantSlug` is now read only inside the `else` arm, so it can never suppress the membership call. The membership check is unconditional on a resolved tenant, which is the invariant the exploit needed broken.

### Availability — the part I was most worried about

The new `else if` hard-fails a login, so the question was whether any legitimate deployment produces "state names a tenant, none resolves". The review enumerated nine configurations and found none that breaks.

The load-bearing detail: a state only carries a slug if the login was served on `{slug}.{basedomain}` and that tenant resolved, because the apex passthrough in `TenantResolutionMiddleware` precedes the sole-tenant auto-resolve. So single-tenant with no `BASE_DOMAIN`, single-tenant on the apex, and proxies rewriting `Host` all produce no slug and land in row 1, unchanged. Proxy setups are safe because both middlewares derive the host identically on login and callback, so the crossed state isn't reachable that way. Multi-replica is fine because the Data Protection key ring is DB-persisted.

The setup callback was the highest-risk candidate and it is genuinely out of reach: `/api/v4/setup/oidc/callback` always carries a slug and never resolves a tenant, but it routes to `HandleSetupCallbackAsync`, which issues its own session and never consults `currentTenantId` — it does not reach `CompleteLoginAsync` at all. The link callback likewise routes to `AttachVerifiedIdentityAsync`.

One behaviour change, benign: if a tenant is deleted inside the ~15-minute state lifetime, that login now returns `invalid_state` instead of issuing a tenantless session for a tenant that no longer exists. The user re-initiates. Not a legitimate flow.

### One test was partially vacuous — fixed in 65069ac9a

`TryReadTenantSlug_ForAStateThisInstanceDidNotIssue_ReturnsNull` passed against a plain base64url decoder as well as the protected one, because `ForgeState` built a payload with no `TenantSlug` — nothing to find either way. Impact was nil (the only consumer regex-gates the slug and can only build a subdomain of the instance's own base domain), but a test that cannot fail is worth less than no test.

The forged payload now names a tenant. Reinstating the old plain decoder in `TryReadTenantSlug` turns 0 failures into 2, so it discriminates. No production change.

The other three new guards were already confirmed non-vacuous, with the review identifying which assertion does the work in each — `IsAccessDenied` for the membership denial (it distinguishes that from the new `invalid_state` denial), and `Error == "invalid_state"` for the hard-fail, a value produced nowhere else in `CompleteLoginAsync`.

### Also confirmed sound

No analogous slug-versus-resolved-tenant confusion elsewhere: `AttachVerifiedIdentityAsync` keys off the caller's authenticated session, and `HandleSetupCallbackAsync` off `Intent`/`SubjectId` — both now inside the protected envelope. Single encode/decode chokepoint with no residual `Base64Url` state handling. Fail-closed on both decode paths. And a detail I had not spotted: moving `ProviderId` inside the protected envelope also closes the ability to swap in an OAuth2 provider to skip the nonce path entirely.

Full API suite: **4180 passed, 0 failed**.

PR #398, opened 2026-06-12
Shared Rust alert engine: golden corpus, nocturne-alerts-core, FFI bindings, backend seam

ryceg:
﻿## Summary

Phases 0?3 of the shared alert-engine work (groundwork for Prelude, the Android companion app): the leaf/node alert evaluation engine is now implemented once in Rust and consumable from both the .NET backend and (next phase) Kotlin via UniFFI ? with the existing C# engine still the default and a corpus-pinned parity guarantee between them.

- **Phase 0 ? semantics freeze**: `docs/alerts/engine-semantics.md` (normative spec incl. anomalies index) + a 103-scenario golden corpus in `tests/Parity/AlertEngineCorpus/`, generated from the live C# engine; `--check` keeps it honest in CI.
- **Phase 1 ? `crates/nocturne-alerts-core`**: pure-Rust port (no I/O, no clock). 103/103 corpus parity + 48 ported unit tests.
- **Phase 2 ? `crates/nocturne-alerts-ffi` + `Nocturne.Core.Alerts.Native`**: JSON-envelope C ABI (catch_unwind, error envelopes) bound via the OrefInterop pattern; 116-test three-way parity suite (C# == corpus == Rust-over-FFI); `native-release.yml` for tagged cdylib builds.
- **Phase 3 ? engine seam**: `IAlertEvaluationEngine` with `Managed`, `RustBacked`, and `Shadow` implementations selected by `Alerts:Engine` (**default `managed` ? production behaviour unchanged**). Shadow mode runs both side-effect-free on the rust path and logs field-level `AlertEngineDivergence` events. Timer-continuity proven in both directions. The 39 C# evaluators are untouched; deletion waits for a shadow soak.

## Test plan
- [x] 687 alert-domain tests (incl. seam corpus tests for both engines) with `NOCTURNE_ALERTS_REQUIRE_NATIVE=1`
- [x] 116 three-way FFI parity tests, 0 skipped
- [x] `cargo test/clippy/fmt` clean (74 tests)
- [x] Corpus checker: 103 scenarios up to date
- [x] Full `Nocturne.API.Tests` suite: 3633 passed
- [x] New `alerts-ffi-parity` + `alert-corpus-check` CI jobs green on this PR (first linux-x64 run)

?? Generated with [Claude Code](https://claude.com/claude-code)

PR #590, opened 2026-07-28
Rust alerts

ryceg:


ryceg:
Superseded by #603.

This branch predated the squash-merge of #398, so main already carries its feature content — the Rust alert-engine port, the golden corpus, the FFI bindings, the `IAlertEvaluationEngine` seam and scoped DND are all on main, and have been evolved further since (#446, #452, the sleep_session_active evaluator, multi-arch cdylib packing).

That is why the PR showed as conflicting with 40 `add/add` conflicts: there was no merge to perform, only a duplicate to retire. Rebasing revealed that main had also independently fixed several findings from the review of this branch, in places better than the branch did — it cross-compiles aarch64 for the multi-arch image, where this branch's single-arch cdylib build would have been a regression.

#603 carries the review-fix remainder that is genuinely not on main (27 files), rebased and green.

PR #605, opened 2026-07-29
refactor(auth): drop the unusable share-token backfill, and cut the narration

ryceg:
Review follow-ups on #597, which is already merged. No change to the at-rest storage formats it introduced.

**`ShareTokenBackfillService` is deleted.** Its purpose was to keep a legacy publicly-readable tenant's public access working once the bare `{slug}` host became login-only, and it cannot do that any more: only the digest is stored, so the token it minted was discarded and no code path — nor an operator with database access — could ever produce the URL. Its whole remaining effect was to set `share_token` to an unresolvable digest, which makes the settings card report public access as **on** for a link nothing can resolve.

Reducing it to "notify without minting" is not available either: `CreateNotificationAsync` does not dedupe on `(type, sourceId)`, so it would re-file on every startup. The tenant's public-read capability lives in its Public subject's roles and is untouched; `share_token = null` is the honest state and already routes the owner to "public access is off, regenerate to re-enable". #597 verified this service matches **0** tenants in production.

**The notifier is renamed** to `IShareLinkRotatedNotifier` / `sharing.link_rotated`, now that the rotation pass is its only caller — matching the verb this area already uses (`RotateAsync`, `RotatePlaintextShareTokensAsync`) instead of adding "reset" beside it. It also fixes the copy: "the previous link no longer opens your data" was false on the backfill path, where there was never a previous link.

**`TotpSecretProtection.RequireProtector`** for the runtime path, used by the startup pass. `CreateProtector` stays lenient because the EF model has to build in design-time and test harnesses that never register Data Protection — making it strict fails eight `ShareTokenCacheServiceTests`, which wire a context to a service provider without it. The pass is the path that matters: writing the column under the ephemeral fallback would leave payloads no later process can read.

**`CredentialHash.Sha256Hex` is private.** The two credentials normalize case in opposite directions before hashing, so the normalization belongs inside the primitive; `GuestLinkService`'s local wrapper is replaced by `CredentialHash.GuestCode` (byte-identical digest, no re-hash needed).

### Tests

The catch in `VerifyLoginAsync` is now asserted rather than explained. #597 claimed no in-process test could produce an undecryptable payload; that was wrong — `CredentialAtRestPassTests` already seeds a raw-plaintext `secret_key` over SQL, and materializing that row through the converter is exactly the failure. Two new integration tests pin the two behaviours the fix depends on:

- the read throws `CryptographicException` **unwrapped** (EF does not wrap converter exceptions, which is what `catch (CryptographicException)` relies on);
- the key-only projection plus stub delete removes such a row, so re-enrolment stays reachable.

Also: `TotpSecretProtectionTests` asserted `Throw<Exception>`, which would pass for a failure the production catch misses — narrowed to `CryptographicException`. `Generate_fits_a_dns_label` asserted `length <= 63` against a hardcoded 16 that the test above already pins, so it could not fail independently — removed. Both pass tests asserted table-global counts, which held only because the xUnit collection serializes — scoped to the seeded row.

### Comments

Cut back to what the code needs: the 17-line block on the TOTP try/catch, the rejected-alternatives paragraphs, and the commentary on which CI jobs run. That rationale lives in the commit history. Also fixes the `using` order in `Program.cs` and drops the redundant logger argument now that `RunAsync` already receives the service provider.

### Verified locally

Full solution builds. `Nocturne.API.Tests` 4366 passed / 5 skipped (pre-existing Rust-engine skips) / 0 failed. `CredentialAtRestPassTests` 9/9 against real PostgreSQL, including both new tests.

### One thing for the operator

`sharing.link_rotated` is a persisted notification type. Safe here because the rotation pass had not yet run anywhere, so no `sharing.link_reset` rows exist. If any environment has already booted #597's image, its existing rows would lose their template and frontend label.


ryceg:
Updated after checking the deploy state, because #597 landed and shipped while this was being prepared.

`build-and-push` for `154838c85` **succeeded** (only `publish-glucose-chart` failed in that run), and the VPS image-poll cron runs every 15 minutes, so the rotation pass has very likely already run in production and filed `sharing.link_reset` notifications for the 4 affected tenant owners.

`resolveNotificationLabel` returns an unmapped key unchanged, so renaming the keys on its own would have shown those owners a literal `share_link_reset`. `d7f350a63` keeps the legacy keys mapped.

The backend side needs no shim: `GetTemplate` is only called from `CreateNotificationAsync`, and each row persists its own category, urgency, icon and source, so retiring the `sharing.link_reset` template does not affect rows already filed under it. Verified by reading the read path, not assumed.

I could not confirm the row count directly — prod database access is not available in this session. If you can check, `select type, count(*) from in_app_notifications where type like 'sharing.%' group by type` settles it; if the pass has **not** run anywhere, `d7f350a63` is harmless dead entries you may prefer to drop.

PR #619, opened 2026-08-02
fix(web): pin one svelte across the workspace so Snippet types stop conflicting

ryceg:
## Problem

The `svelte` override in `src/Web/pnpm-workspace.yaml` was a **range** (`^5.37.0`). A range override doesn't force a single version — it only rewrites each package's requested range, so pnpm is free to keep several satisfying copies. This tree had three installed at once:

```
svelte@5.45.2
svelte@5.50.0
svelte@5.55.5
```

Svelte's `Snippet` is branded with a `unique symbol` declared **per copy**, so as soon as two copies land in one type-check program, every snippet-to-snippet assignment fails:

```
Type '() => ReturnType<import("svelte").Snippet>' is not assignable to type 'Snippet<[]>'.
  ... Two different types with this name exist, but they are unrelated.
```

That accounts for **20 errors** across `packages/glucose-chart/src/dialogs/` (`DeliveryInspectionDialog`, `GlucoseInspectionDialog`, `TreatmentInspectionDialog`). There is nothing wrong with that code — it's purely the duplicate copies.

## Fix

Pin the override to an exact version. Verified with `svelte-check` against `packages/glucose-chart` before and after: **20 errors in `src/` → 0**, and the lockfile drops to a single `svelte@5.55.5` entry (net −200 lines).

`5.55.5` is what `app` and `desktop` already request, and it satisfies `portal`'s `^5.45.4` plus the `>=5.0.0` peer ranges in the library packages — nothing crosses a major.

## Why it matters beyond this repo

A `link:`/path-based consumer resolves svelte from the linked package's *own* `node_modules`, so a second copy leaks into the consumer's program and the identical 20 errors surface there. We hit exactly this consuming `@nightscout/glucose-chart` from the submodule in nocturne-cloud.

## Verification

- `pnpm install` clean; `svelte-check` on `packages/glucose-chart`: 20 → 0 errors in `src/`.
- I could **not** run `pnpm --recursive check` locally: `svelte-check-native` requires `@typescript/native-preview` (tsgo), which isn't installed in this tree, and `packages/app`'s check additionally needs the generated API client (a `dotnet build` first). Please let CI/reviewers confirm the app and portal builds — that's the main risk in an exact pin.

## Two gaps I deliberately left alone

1. **`glucose-chart` is the only Svelte package with no `check` script**, so root `pnpm --recursive check` silently skips it — which is why these 20 errors sat unnoticed. I drafted the one-liner (`svelte-check-native --tsconfig ./tsconfig.json`, matching `cms`/`coach`/`ui`) but dropped it rather than ship a script I couldn't execute here; none of those packages set `skipLibCheck`, and plain `svelte-check` reports ~52 third-party `.d.ts` errors, so the script likely needs `skipLibCheck: true` alongside it.
2. **No workflow runs `pnpm check`** — the only match in `.github/workflows/` is `--no-git-checks` in a publish step. Adding the script without a CI step wouldn't catch regressions either.

https://claude.ai/code/session_015nep4BaV3eeixbxBXZqXwb

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -148,6 +148,37 @@ jobs:
   # Database integration tests: bare-Postgres Testcontainers (no Aspire), so they
   # run reliably on GitHub-hosted runners where Docker is available. Other
   # integration projects stay deferred until their fixtures/references are fixed.
+  # Typechecks the Svelte library packages that have no build step and no generated API
+  # client, so this needs nothing but pnpm. @nocturne/app and @nocturne/portal are excluded
+  # because their checks depend on the NSwag/Zod codegen and belong with the .NET build.
+  #
+  # Only glucose-chart and coach are listed. @nocturne/ui and @nocturne/cms still run
+  # svelte-check-native, which silently discovers zero files on Linux (see the note on their
+  # `check` scripts) — and once pointed at a checker that does read their sources they report
+  # ~880 and ~838 errors respectively. That debt has to be triaged before either can be gated.
+  web-typecheck:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    defaults:
+      run:
+        working-directory: src/Web
+    steps:
+      - uses: actions/checkout@v4
+      - uses: pnpm/action-setup@v4
+        with:
+          # src/Web's lockfile is v11 format, so pnpm 9 (what the unit job pins) cannot read it
+          # with --frozen-lockfile. Same pin the image build uses in docker-publish.yml.
+          version: "11.5.0"
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "24"
+          cache: pnpm
+          cache-dependency-path: src/Web/pnpm-lock.yaml
+      - run: pnpm install --frozen-lockfile
+      - name: Typecheck library packages
+        run: |
+          pnpm --filter @nightscout/glucose-chart                --filter @nocturne/coach                run check
+
   integration-db:
     runs-on: ubuntu-latest
     timeout-minutes: 20
--- a/src/Web/package.json
+++ b/src/Web/package.json
@@ -22,6 +22,7 @@
   },
   "devDependencies": {
     "@types/node": "^24.10.1",
+    "@typescript/native-preview": "7.0.0-dev.20260513.1",
     "prettier": "^3.7.2",
     "tsx": "^4.20.6",
     "typescript": "^5.9.3",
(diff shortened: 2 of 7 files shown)


ryceg:
Heads up on the red checks: `build-and-push`, `unit`, `alert-corpus-check` and `alerts-ffi-parity` are **failing on `main` too**, not from this PR.

`main`'s own "Build and Publish with Aspire" run fails with the same compile error:

```
error CS0103: The name 'DataSources' does not exist in the current context
  src/Connectors/Nocturne.Connectors.CareLink/Mappers/CareLinkSensorGlucoseMapper.cs(51,30)
```

This PR touches only `src/Web/pnpm-workspace.yaml` and `src/Web/pnpm-lock.yaml`, so it cannot reach the CareLink connector. `integration-db`, `CodeQL` and all four `Analyze` jobs are green here.

Worth noting it means CI can't currently validate the app/portal builds against the exact svelte pin — the thing I most wanted a second pair of eyes on. Happy to rebase once main is green.

ryceg:
Update — I tried to add a CI job for this and **backed it out**, because it doesn't actually check anything.

On the Linux runner, three of the four packages report:

```
packages/glucose-chart check: svelte-check found 0 errors and 0 warnings in 0 files
packages/coach check:         svelte-check found 0 errors and 0 warnings in 0 files
packages/ui check:            svelte-check found 0 errors and 0 warnings in 0 files
```

**Zero files.** The identical commands check 12, 3 and 208 files on Windows. A gate that passes without inspecting anything is worse than no gate — it reads as proof. So the job is out until someone understands why the file discovery comes up empty there.

Two other things that fell out of the attempt, both worth knowing:

1. **`svelte-check-native@0.2.0` mis-tokenizes comments on Linux.** The `// eslint-disable-next-line @typescript-eslint/no-explicit-any` on `ChartLegend.svelte:129` came back as nine errors — `Cannot find name 'no'`, `Cannot find name 'explicit'`, `'any' only refers to a type` — all attributed to line 1. Doesn't reproduce on Windows. `cms`, `coach` and `ui` are still on 0.2.0. I moved glucose-chart to ^0.8.5 (what `app` uses), which also needs `skipLibCheck` since it reads dependency `.d.ts` files and `app` only gets that from SvelteKit's generated tsconfig.

2. **`cms` has 22 pre-existing type errors in 9 files** — it was the one package that did check its sources on Linux. `Cannot find module 'mdast'`, an `onMouseMove` that isn't on the options type, a non-callable `.use(remarkHtml, …)`. Not touched here, but it means cms can't be gated until those are fixed.

What this PR still does: the svelte pin (the actual fix for the 20 glucose-chart errors), the missing `check` script so `pnpm --recursive check` stops silently skipping the package, and `@typescript/native-preview` at the workspace root — without it `pnpm run check` in cms, coach or ui fails with "could not find tsgo" rather than typechecking, so three of those four scripts were dead on arrival.

PR #623, opened 2026-08-02
fix(auth): resolve the legacy read scope from the storage, and gate the v1 read plane

ryceg:
Six commits closing the legacy (v1/v2/v3) read plane, plus two anonymous endpoints.

## What was open

`origin/main` carries **no** `RequireScope` on `TimeQueryController` or `CountController`. Any caller with a non-empty resolved scope set reached `slice/{storage}/…`, `times/…`, `times/echo?storage=` and `count/{storage}/where` for every collection those routes serve, regardless of which categories their grant covered.

## Approach

A class- or action-level scope list is an OR across every collection a route can serve, so it would admit a caller holding one category to all of them. Where the collection is a route or query value it is resolved per request instead, through a new `LegacyStorageReadScopes`; where it is statically known the attribute is correct and is used, so the attribute scan can see it.

- `slice/{storage}/…` and `times/echo?storage=` — per-request resolution in the handler.
- `times/{prefix}` and `times/{prefix}/{regex}` — `[RequireScope(glucose.read)]`; they take no selector, `GetTimeBasedEntriesInternal` hardcodes `entries`.
- `count/{storage}/where` — same resolution, except `activity`, which merges four categories into one unfilterable number and so requires all four.
- An unclassified selector is refused rather than passed through, so adding a collection to a route without classifying it closes the route rather than opening it. `CountController`'s allow-list is asserted fully classified by a test.

Also: `isInternalOnlyApiPath` did not percent-decode, so `/api/v4/platform/%74ls-authorize` reached an endpoint the proxy is supposed to make internal-only. It now decodes per segment.

## Verification

Every gate was proven non-vacuous by reverting it and confirming failure — removing the three `times` attributes fails 6 tests; removing the decoding fails 5 vitest cases; adding an unclassified selector to the count allow-list fails the classification guard.

4548 API tests pass. Two failures are pre-existing and unrelated: `CacheIntegrationTests.CreateEntriesAsync_DecomposesToV4AndFiresEvents` (reproduced on the unmodified branch) and `ConnectorCursorResetJobServiceTests` (passes in isolation; the branch touches no connector files).

## Reviewed

An independent review of the final commit found the `%2F` doc claim was false (`.join("/")` refolds it), an orphaned XML doc introducing a `CS1587` warning, `CanRead` left production-dead while `RefuseRead` reimplemented it, and an unannounced change to `CountController`'s 400 body. All are fixed here. Note that rows added to `V1AuthenticationRegressionTests` pin the anonymous bare-host contract — they pass with or without the scope gate, and are not the regression test for it; that is `TimeQueryStorageGateTests`.

## Known, not addressed here

`GET /api/v1/times/echo?storage=profile` passes the gate and then 500s out of `TimeQueryService`, which advertises 5 storages to a route serving 3. Pre-existing on main.

https://claude.ai/code/session_01Wk3jH4N16htaBqGod2j7y8

### DIFF
--- a/src/API/Nocturne.API/Attributes/RequireScopeAttribute.cs
+++ b/src/API/Nocturne.API/Attributes/RequireScopeAttribute.cs
@@ -17,6 +17,11 @@ namespace Nocturne.API.Attributes;
 /// hierarchical scope matching (e.g., <c>read</c> satisfies <c>read:entries</c>).
 /// The granted scopes are further refined by <see cref="Middleware.MemberScopeMiddleware"/>
 /// based on the user's tenant membership roles.
+/// <para>
+/// The check keys off the resolved scope set rather than an authenticated identity, so it also
+/// applies to the anonymous public-share principal, whose scopes are narrowed to
+/// <see cref="TenantPermissions.PublicShareScopes"/>. An empty scope set is rejected.
+/// </para>
 /// </remarks>
 /// <seealso cref="RequirePermissionAttribute"/>
 /// <seealso cref="Middleware.AuthenticationMiddleware"/>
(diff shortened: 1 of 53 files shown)


ryceg:
Second review pass verified the fixes to the first review's 9 findings: 8 FIXED, 1 PARTIAL (a duplicated test row), plus 2 new defects it found in the amendment. All three are addressed in `60c752c0c`:

- **The new classification test carried a false rationale** — a repeat of the very finding it shipped alongside. An unclassified selector is refused by `RefuseRead`, not passed through ungated. Comment corrected.
- **The `slice` 400-body change survived the `count` fix.** `slice` and `times/echo` dispatch on three collections while the scope map classifies five, so `profile` and `food` cleared the gate on a therapy/food grant and then hit a service that does not serve them — 403 for one caller, 400 for another, on a route that never served the collection. `times/echo` 500'd out of `TimeQueryService` for the same reason. Both now check the dispatch list ahead of the scope gate, restoring the exact 400 body `slice` returned on main and closing the echo 500.
- Duplicated vitest row removed.

Non-vacuity proven by reverting: dropping the pre-gate check and adding an unclassified selector to the list fails 3 tests. 4549 API tests pass; the 2 failures are the known pre-existing `CacheIntegrationTests` and the memory benchmark. 62 vitest pass.

The review's remaining note — `CountableStorage`/`SliceableStorage` being mutable arrays rather than `FrozenSet` — is left as-is; no mutator exists and the surrounding code uses plain arrays throughout.

PR #626, opened 2026-08-02
fix(nightscout): stream backfill pages instead of accumulating full history

ryceg:
Follow-up to #625, which made full-history backfills actually run — and immediately exposed that the connector holds the entire fetched range in memory before publishing. On the first real production migration, the device-status phase fetched **485,000** documents into one list and the publish threw `OutOfMemoryException` inside a 4G container, transiently failing *other tenants'* connector publishes under the same memory pressure. Glucose and treatments survived only because their documents are small.

## Change

- The four near-identical pagination loops (entries, treatments, devicestatus, activity) collapse into one shared `FetchPagesAsync<T>` newest-first page iterator (same bounds/termination semantics as before, including #625's unbounded-fetch anchoring).
- `PerformSyncInternalAsync` consumes pages with `await foreach` and publishes each page as it lands, so peak memory is one page (`MaxCount`, default 1000) regardless of history depth. Counts and `LastEntryTimes` accumulate across pages; the per-type "Retrieved N ... from Nightscout" summary log is preserved (ops greps rely on it).
- A page publish failure **stops the crawl** rather than continuing to hammer the source while the local store is failing. Pages arrive newest-first, so the un-published remainder is older history — a re-run (or admin cursor reset) re-pulls it idempotently. This matches the pre-existing failure profile: partial newest-first publishes already left older gaps; now we just stop paying for pages we can't store.
- `FetchGlucoseDataRangeAsync` / `FetchTreatmentsAsync` remain as accumulating wrappers over the same iterator for external callers (`Tools.Connect`, explicit-range fetches).

## Tests

- New: `SyncData_PublishesEachPageBeforeFetchingTheNext` — with every publish rejected and two full pages queued, exactly one data page is fetched. This pins the interleaving; the old fetch-everything-first code fetches both pages and fails it.
- All 19 existing pagination/retry tests pass unchanged; full `Nocturne.API.Tests`: 4545 passed.

### DIFF
--- a/src/Connectors/Nocturne.Connectors.Nightscout/Services/NightscoutConnectorService.cs
+++ b/src/Connectors/Nocturne.Connectors.Nightscout/Services/NightscoutConnectorService.cs
@@ -196,23 +196,46 @@ public class NightscoutConnectorServiceBase<TConfig> : BaseConnectorService<TCon
         // Handle Glucose
         // Glucose keeps request.From — for background syncs the framework already derived
         // it from the latest glucose entry, so it is glucose's own independent cursor.
+        //
+        // Each data type below streams fetch-page → publish-page rather than accumulating
+        // the whole range first: a multi-year backfill of a high-volume collection held in
+        // one list has taken the process out with OutOfMemory, failing unrelated tenants'
+        // publishes with it. A page publish failure stops the crawl — pages arrive newest
+        // first, so later pages are older history that a re-run re-pulls (idempotent).
         if (activeTypes.Contains(SyncDataType.Glucose))
         {
             try
             {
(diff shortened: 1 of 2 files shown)


ryceg:
Review follow-up pushed (bbfe7792b): a failed page publish no longer breaks the crawl — every page is still attempted, so holes stay bounded to the failed pages (matching the pre-streaming batch-level blast radius) instead of stranding all older history below the advancing cursor. Remaining known gap, deliberately deferred: a fetch exception mid-crawl still strands the older unfetched pages below the cursor (pre-streaming, a fetch failure published nothing and self-healed). Healing that properly needs a persisted backfill low-water mark so an interrupted initial backfill resumes — follow-up material, pairs naturally with #627's job records.

PR #627, opened 2026-08-02
feat(jobs): persist migration and cursor-reset job records across restarts

ryceg:
Companion to #625/#626, fixing the third failure mode from the same production migration: **background job state lived only in memory**.

## What that cost in practice

- The tenant's onboarding-wizard import ran on Aug 1; by the time we investigated, the API had restarted and there was **no evidence the import had ever run** — no job record, no history, nothing to distinguish "wizard never started it" from "it ran and died".
- The first recovery cursor-reset imported 253k readings, then the redeploy cron restarted the container: the poll went from `Running` to a **bare 404**, and the work silently died with it.
- `migration_runs` / `migration_sources` tables have existed in the schema all along but were **never written** — `GET /api/v4/migration/sources` always returned `[]`, and `MigrationStartupService`'s "has this source already been migrated?" check could never be satisfied, making the `MIGRATION_MODE` startup notification fire forever.

## Change

- **Records are persisted at every lifecycle transition** — migrations into the existing (previously write-orphaned) `migration_runs`/`migration_sources`, cursor resets into a new `connector_reset_jobs` table. The initial write happens before the detached task starts and propagates failure (a job whose record can't be written must not start); subsequent writes log instead of throwing, and the terminal snapshot in the `finally` retries.
- **Status/cancel fall back to the persisted record** when the job isn't in the in-memory map; history and sources are served from the database. Cancel of a persisted-but-not-running job is a logged no-op, not a 404.
- **Startup sweep marks orphans `Interrupted`** (new state on both job enums, with an explicit "re-run it" error message): the detached tasks die with the process, so a `Pending`/`Running` row at boot can only be an orphan. Runs under the migrator role alongside migrations/RLS reconciliation.
- `migration_runs` gains `tenant_id`, `mode`, `source_description`, `created_at` (EF migration `DurableJobRecords`). Lookups are tenant-scoped by query. Both tables are deliberately **not** RLS/`ITenantScoped`: they're operator metadata read cross-tenant (platform-admin endpoints, the startup sweep) and carry lifecycle state only, no health data. Only non-usable SHA-256 digests of source credentials are stored — never the API secret or connection string, and not the SHA-1 form Nightscout accepts as a credential.
- `IConnectorCursorResetJobService.GetStatus/Cancel` became async for the DB fallback; the admin controller endpoints follow.

## Deliberately out of scope

Resuming interrupted work automatically. Re-ingest is idempotent, so the honest and safe semantic is "Interrupted — re-run it": auto-resuming heavy crawls on every boot risks a stampede after a bad deploy.

## Tests

- Migration + reset job services: existing ownership/lifecycle tests updated to the persisted world (they now run against an InMemory `NocturneDbContext`), plus new regressions: status/history served from the persisted record by a fresh service instance over the same store (models the post-restart process), `GetSourcesAsync` returns the upserted source, cancel of a persisted terminal job doesn't 404.
- Full `Nocturne.API.Tests`: 4548 passed; `Nocturne.Infrastructure.Data.Tests`: 481 passed.

### DIFF
--- a/src/API/Nocturne.API/Controllers/V4/PlatformAdmin/ConnectorAdminController.cs
+++ b/src/API/Nocturne.API/Controllers/V4/PlatformAdmin/ConnectorAdminController.cs
@@ -104,11 +104,11 @@ public class ConnectorAdminController : ControllerBase
     [RemoteQuery]
     [ProducesResponseType(typeof(ConnectorResetJobStatus), StatusCodes.Status200OK)]
     [ProducesResponseType(StatusCodes.Status404NotFound)]
-    public ActionResult<ConnectorResetJobStatus> GetResetJobStatus(Guid jobId)
+    public async Task<ActionResult<ConnectorResetJobStatus>> GetResetJobStatus(Guid jobId, CancellationToken ct)
     {
         try
         {
-            return Ok(_resetJobService.GetStatus(jobId));
+            return Ok(await _resetJobService.GetStatusAsync(jobId, ct));
         }
         catch (KeyNotFoundException)
         {
@@ -125,11 +125,11 @@ public class ConnectorAdminController : ControllerBase
     [RemoteCommand(Invalidates = ["GetResetJobStatus"])]
     [ProducesResponseType(StatusCodes.Status204NoContent)]
     [ProducesResponseType(StatusCodes.Status404NotFound)]
-    public IActionResult CancelResetJob(Guid jobId)
+    public async Task<IActionResult> CancelResetJob(Guid jobId, CancellationToken ct)
     {
         try
         {
-            _resetJobService.Cancel(jobId);
+            await _resetJobService.CancelAsync(jobId, ct);
             return NoContent();
         }
         catch (KeyNotFoundException)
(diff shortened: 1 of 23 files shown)


ryceg:
Rebased onto main (the branch previously stacked on #626's commit — it now contains only the durable-jobs work) and pushed review fixes (7cd63be1e):

- **Blocking finding fixed**: `migration_sources` is now tenant-scoped — dedupe and the unique index moved to `(tenant_id, source_identifier)`, and `GET /migration/sources` returns only the calling tenant's sources. Previously (latent until this PR made the table live) any tenant admin could list every tenant's Nightscout URLs.
- The startup pending-migration check now derives identifiers from the same helpers as the upsert (trailing-slash-trimmed URL / SHA-256 of the Mongo connection string), so it can actually match.
- Cancel no longer rewrites a persisted terminal outcome; date ranges normalize to UTC; jobs register in memory only after the record persists; sweep documents its single-instance assumption; the web app treats Interrupted as terminal and badges it.

Full API suite 4547 passed, infra suite 481 passed.

PR #632, opened 2026-08-03
fix(glucose): maintain tenants.last_reading_at at the write chokepoint

ryceg:
Eight active production tenants have glucose data but `tenants.last_reading_at IS NULL` — because the only writer was `GlucosePublisher` (the connector publish path). Tenants whose glucose arrives via **direct uploader POSTs** (Trio, xDrip, AAPS via v1) never got it set. The column is load-bearing: `AlertSweepService` uses it as the staleness signal (`LastReadingAt ?? DateTime.MinValue`), so these tenants permanently read as "no data ever" to the sweep, and `TenantOverviewService` shows their glucose status from the same column.

## Change

- The update moves into `SensorGlucoseRepository` (`CreateAsync` + a `BulkCreateAsync` override) — the one chokepoint every sensor glucose write passes through: uploader POSTs (via the entry decomposer), connector publishes, and imports. The publisher's private copy is deleted along with its now-unneeded `IDbContextFactory`/`ITenantAccessor` dependencies.
- Semantics tightened from *ingestion wall-clock* to *newest reading timestamp, advance-only* (`WHERE last_reading_at IS NULL OR last_reading_at < newest`). Rationale: the canonical alert evaluator already derives recency from `latest.Timestamp`, and the old behavior let a backfill of years-old history stamp `now` — making a silent sensor look fresh and suppressing real staleness alerts during imports. Meter-only (fingerstick) publishes no longer bump it; CGM staleness is what the readers measure.
- Failure to advance is logged, never thrown — the reading write matters more than the recency column.

## Healing

No data migration: affected tenants self-heal on their next reading (minutes, for the live-uploader tenants). Tenants with no active source keep NULL, which is the honest value.

## Tests

New `SensorGlucoseRepositoryLastReadingTests` (SQLite harness): bulk create advances to the newest timestamp of the batch, backfill/out-of-order batches never move it backwards, single create advances. Full API suite 4548 passed, infra suite 484 passed.

### DIFF
--- a/src/API/Nocturne.API/Services/ConnectorPublishing/GlucosePublisher.cs
+++ b/src/API/Nocturne.API/Services/ConnectorPublishing/GlucosePublisher.cs
@@ -26,8 +26,6 @@ internal sealed class GlucosePublisher : IGlucosePublisher
     private readonly ISensorGlucoseRepository _sensorGlucoseRepository;
     private readonly IMeterGlucoseRepository _meterGlucoseRepository;
     private readonly IPatientDeviceStamper _patientDeviceStamper;
-    private readonly IDbContextFactory<NocturneDbContext> _contextFactory;
-    private readonly ITenantAccessor _tenantAccessor;
     private readonly ICanonicalAlertEvaluator _alertEvaluator;
     private readonly IAuditContext _auditContext;
     private readonly ILogger<GlucosePublisher> _logger;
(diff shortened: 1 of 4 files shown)


ryceg:
Review follow-up pushed (87d6a3c7d): the advance now clamps to `UtcNow` — advance-only semantics would otherwise let a single future-dated reading (device clock skew / double-applied tz offset, both documented in the wild here) pin `last_reading_at` ahead of now permanently, silently disabling the staleness and signal-loss evaluators that compare `now - LastReadingAt`. New tests cover the clamp and the SyncId-upsert arrival path. Note from the review worth recording: meter-only (fingerstick) publishes no longer bump the column — every reader treats it as CGM recency, so the old meter bump suppressing signal-loss was itself a bug; a hypothetical meter-only tenant now honestly reads as having no CGM data.

PR #633, opened 2026-08-03
feat(nightscout): persist a backfill low-water mark and resume below it

ryceg:
The resumability follow-up deferred from #626. Pages stream newest-first and every resume cursor derives from the latest *stored* record, so a backfill crawl killed mid-run — publish failure, fetch failure, or a process restart (the redeploy cron restarts the API routinely; it killed a real production backfill twice in one day) — stranded everything older than its last published page below the cursor, permanently invisible to catch-up syncs.

## Mechanism

- **A per-collection low-water mark** persists on `connector_configurations` (new `backfill_low_water_marks` jsonb — runtime sync state alongside `last_successful_sync`, never part of the user-editable configuration, so a config PUT can't clobber it). Connectors reach it through two new `IMetadataPublisher` members; the API-side implementation resolves the config row from the connector source (`nightscout-connector` → `nightscout`).
- **Full crawls (open lower bound) advance the mark after every published page** — crash protection for multi-hour histories — and **clear it on reaching the source's beginning**. The 3.5-hour, 533k-document device-status re-pull we ran yesterday would have resumed instead of restarting from scratch had it been killed.
- **On the next sync, an existing mark triggers a bounded resume crawl** (`…find[date][$lte]=mark−1ms`) after the ordinary catch-up — skipped when the primary crawl already failed, so a failing store isn't hammered with deep history too.
- **A page publish failure stops the crawl again** (reverting #626's attempt-every-page): with resume, stopping is now the correct behavior — older pages are no longer stranded, and holes never scatter (the mark describes one contiguous "missing below X" region). Bounded catch-up crawls only ever *raise* the mark, never lower or clear it (a deeper mark from an earlier failure still describes missing history further down).
- The four per-type sync blocks collapse into one `CrawlAndPublishAsync` helper (count/newest-time/publish/mark logic was 4× duplicated).

## Also fixes: silent truncation on fetch failure

`FetchDataAsync` reports failure (retries exhausted, non-retryable HTTP, bad JSON) as `null`, which the pagination loop treated as a clean end-of-range — a partial sync reported **successful** (flagged as pre-existing in #626's review). The mark machinery would have amplified this into wrongly *clearing* marks. A null page now fails the collection's sync; a genuinely exhausted range answers `[]`. Connector health goes unhealthy and the next cycle retries, instead of quiet data loss.

## Tests

New `NightscoutBackfillResumeTests` (in-memory mark store over the `IMetadataPublisher` seam):
- publish failure records the mark at the bottom of the last published page and stops the crawl
- fetch failure mid-crawl records the mark and fails the sync (regression for the silent-truncation hole)
- a completed full crawl clears a stale mark
- an existing mark triggers exactly one resume crawl bounded just below it, counts both regions, and clears on completion
- no mark → no resume crawl

`SequentialMockHandler` extracted from the pagination tests for reuse. Full `Nocturne.API.Tests`: 4553 passed; Connectors Core 62, Infrastructure.Data 481 passed. EF migration `BackfillLowWaterMarks` adds the column.

## Scope

Nightscout connector only — it is the one connector doing unbounded full-history crawls (`InitialSyncFloor => null`). Other connectors' 6-month floors bound the same failure shape; generalizing the helper into `BaseConnectorService` is a follow-up if it ever bites there.

### DIFF
--- a/src/API/Nocturne.API/Services/ConnectorPublishing/MetadataPublisher.cs
+++ b/src/API/Nocturne.API/Services/ConnectorPublishing/MetadataPublisher.cs
@@ -301,6 +301,14 @@ internal sealed class MetadataPublisher : IMetadataPublisher
     }
 
     /// <inheritdoc />
+    /// <remarks>
+    /// The update is a single jsonb-path statement on PostgreSQL so concurrent writers to
+    /// DIFFERENT collections on the same row (a manual sync or cursor-reset job racing the
+    /// background sync) can't clobber each other's keys — a read-modify-write of the whole map
+    /// could drop a mark, and a dropped mark is stranded history, the exact failure marks exist
+    /// to prevent. Same-key races stay last-writer-wins: any surviving mark is a valid resume
+    /// point because the resume crawl is unbounded below it.
+    /// </remarks>
     public async Task SetBackfillLowWaterMarkAsync(
         string source,
         string collection,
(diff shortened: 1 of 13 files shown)


ryceg:
Review follow-ups pushed: mark writes are atomic per-key jsonb-path updates on PostgreSQL (the whole-map read-modify-write could drop another collection's mark when a manual sync or reset job raced the background sync — and a dropped mark is stranded history again); the deliberately-unbounded resume is now documented as load-bearing on CrawlAndPublishAsync, including the accepted rare cost of a full re-crawl after a failed bounded catch-up. Test gaps closed: failed-primary-skips-resume-and-preserves-mark, bounded-failure-never-lowers-a-deeper-mark, and real MetadataPublisher round-trip coverage (source-name mapping, per-collection independence, clear-to-null, unknown-source no-op); the fake mark store now binds to the exact source string so a mapping bug can't hide. Suites: 4559 API / 481 infra green.

PR #648, opened 2026-08-06
refactor(config): derive the public origin from BASE_DOMAIN everywhere

ryceg:
## Problem

A self-hoster reported that the generic OIDC provider always sends `redirect_uri=http://localhost:5000/api/auth/oidc/callback` no matter what `BASE_DOMAIN` is set to.

Root cause: the public origin was configurable two ways. `BASE_DOMAIN` (required; drives tenant resolution and the WebAuthn RP ID/origin) and a separate top-level `BaseUrl` key read by OIDC redirects, post-logout redirects, return-URL validation, the well-known issuer, Pushover callbacks, and invite links. Nothing in the self-hosting path ever set `BaseUrl` — the compose bundle and the Aspire host only set `BASE_DOMAIN` — so `OidcAuthService.GetRedirectUri` always fell through to its hardcoded `http://localhost:5000`. (The Helm chart and the cloud wrapper set `BaseUrl` explicitly, which is why hosted deployments never hit this.)

## Change

`BASE_DOMAIN` is now the single source of truth. `BaseDomainOptions` gains a `PublicOrigin` property (`https://{BaseDomain}`; HTTPS is already a platform-wide assumption via the WebAuthn origin derivation, and both bundled edges — Caddy and Cloudflare — terminate TLS) and every former `BaseUrl` consumer reads it:

- `OidcAuthService` (redirect_uri + post_logout_redirect_uri) — now **throws** when `BASE_DOMAIN` is missing instead of silently sending localhost to the provider
- `OidcController.IsValidReturnUrl` — also tightened to match on the origin *boundary*; the old prefix check accepted `https://example.com.evil.com`
- `WellKnownController` (issuer; request-host fallback unchanged)
- `ServicesController` (apex fallback for uploader URLs)
- `PushoverService` — emergency-receipt callbacks are now omitted when no public origin is configured, rather than pointed at localhost
- `MemberInviteService` (apex fallback for invite links)

Deleted dead config surface: `ServiceNames.ConfigKeys.BaseUrl`, the never-read `Oidc:BaseUrl` property, and the `Program.cs` bridge filling `NocturneApiUrl` (which has no reader anywhere).

## Breaking changes

- The `BaseUrl` env var / config key is **removed** (it was undocumented for self-hosters; anyone setting it should set `BASE_DOMAIN` to the bare host instead).
- **Helm:** `values.baseUrl` → `values.baseDomain` (bare host, no scheme). Templates `fail` fast with a migration message if the legacy value is set. This also fixes the web container, whose `BASE_DOMAIN` env previously received a full URL its consumers don't expect. Chart bumped to `0.1.0-alpha.10`.
- The compose bundle needs **no change** — it already sets only `BASE_DOMAIN`, and simply starts working.

## Tests

- New `OidcAuthServiceRedirectUriTests`: `redirect_uri` is `https://{BASE_DOMAIN}/api/auth/oidc/callback`, and URL generation fails loudly without a base domain.
- Existing OIDC/invite/Pushover/ChatBot test fixtures migrated from mocked `BaseUrl` config to `BaseDomainOptions`.
- Full `Nocturne.API.Tests` run: green except two pre-existing local-only failures unrelated to this change (`CacheIntegrationTests` is `Category=Integration` so CI never runs it, and it has gone stale against the `DecomposeBatchAsync` switch; plus a memory benchmark that's environment-sensitive).

## Follow-ups (not in this PR)

- nocturne-cloud: drop the now-ignored `nocturneApi.WithEnvironment("BaseUrl", ...)` from its AppHost at the next submodule bump.
- `CacheIntegrationTests` staleness above deserves its own fix.

https://claude.ai/code/session_01Czjkxvbn2BbPmU3ideUJe4

### DIFF
--- a/deploy/helm/nocturne/Chart.yaml
+++ b/deploy/helm/nocturne/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: nocturne
 description: Nocturne — modern Nightscout for diabetes data, on Kubernetes.
 type: application
-version: 0.1.0-alpha.9
+version: 0.1.0-alpha.10
 appVersion: "latest"
 kubeVersion: ">=1.27.0-0"
 home: https://github.com/nightscout/nocturne
--- a/deploy/helm/nocturne/README.md
+++ b/deploy/helm/nocturne/README.md
@@ -79,7 +79,7 @@ kubectl create secret generic nocturne-db-web \
 
 # 2. Write a values file.
 cat > my-values.yaml <<EOF
-baseUrl: https://nocturne.example.com
+baseDomain: nocturne.example.com
 
 instanceKey:
   existingSecret: nocturne-instance-key
@@ -157,7 +157,7 @@ The full set of configurable values is in [`values.yaml`](./values.yaml). Highli
 
 | Key | Description |
 |---|---|
-| `baseUrl` | Public URL the deployment is reachable at. Used by the API for OIDC redirects, invite links, etc. |
+| `baseDomain` | Public base domain the deployment is reachable at (bare host, no scheme; HTTPS is assumed at the edge). Drives tenant subdomains, the WebAuthn RP ID, OIDC redirects, and invite links. |
 | `instanceKey.existingSecret` | Secret containing the shared HMAC key. **Required.** |
 | `externalDatabase.host` / `.port` / `.database` / `.sslMode` | Postgres connection details. |
 | `externalDatabase.{app,migrator,web}Secret.existingSecret` | Secret with each role's password under key `password` (override with `existingSecretKey`). |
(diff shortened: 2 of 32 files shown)


ryceg:
Post-review hardening pushed in 3015dada1. A review pass on this PR's own return-URL change found that the relative-URL branch still accepted protocol-relative values: `Uri.TryCreate("//evil.com", UriKind.Relative)` succeeds, and browsers resolve `Location: //evil.com` off-site — so the advertised open-redirect fix was only half-delivered.

`IsValidReturnUrl` now:
- accepts site-local paths only if they start with `/` but not `//` or `/\`
- parses absolute URLs and compares scheme + authority against the public origin (so `https://{origin}?query` works, while `https://{origin}.evil.com`, `https://{origin}@evil.com`, and userinfo on the real origin do not)

New `OidcControllerReturnUrlTests` covers all of the above through the login endpoint.

Two deliberate non-changes from the same review, for the record:
- Path-relative return URLs without a leading slash (`settings/account`) are now rejected; every caller passes absolute paths, and accepting them would make validity depend on the current request path.
- A `BASE_DOMAIN` mistakenly set with a scheme (`https://example.com`) still isn't specially validated — it already breaks tenant resolution and WebAuthn long before OIDC, and the Helm template fails fast on the migrated value.

https://claude.ai/code/session_01Czjkxvbn2BbPmU3ideUJe4

PR #649, opened 2026-08-06
fix(share): make the report read surface reachable by public share links

### LINKED ISSUE #635: Public sharable link not allowed to view most of the data
I made a public link for my doctor with all permissions, but the sharable link is not allowed to view most of the data.
A lot of views throw a 401 inside the reports menu:

Image
Image

But some do show data:

Image
Image

The configuration of my link:

Image

(I can share my public link via DM is needed)
Also the sharable link is not shown anymore after the API has updated. The link does still work, just not shown

ryceg:
Fixes #635.

## What was broken

A public share link granted every category still threw 401 on most report views. Two independent causes:

1. **Bare `[Authorize]` on the report-feeding V4 controllers.** The share principal is deliberately `IsAuthenticated: false` — its access is supposed to come from the `HasPermissions` fallback policy plus the per-category share RLS. `[Authorize]` therefore 401'd every share regardless of granted categories. The views that *did* work (year overview, chart-backed pages) are served by the attribute-less controllers in `FallbackGatedControllers` (ChartData, DataOverview, Actogram, Retrospective, Prediction).
2. **`resolveReportRange` reads the patient timezone from the profile summary** (therapy settings — intentionally not a shareable category), so every report page failed on that call even where its data endpoints were reachable.

The other half of the issue — "the sharable link is not shown anymore" — is the credential-at-rest design working as intended: only the token digest is stored, so the URL is displayable exactly once, at creation/regeneration. The sharing page copy already explains this.

## The fix

- Class-level `[Authorize]` → class-level `[RequireScope(<category>.read)]` on the glucose (sensor/meter/bg-check/calibration), treatments (bolus/nutrition/basal-injection/bolus-calculation/note) and device-status (device-event/aps/pump/uploader/battery) families plus SensorIntegrity; per-action on Statistics (`reports.read` for report analytics, matching the "Reports" share category; `glucose.read` for `periods` and `punch-card`, which the dashboard widgets and calendar render). TempBasal, HeartRate and StepCount already had complete per-action gates and only lose `[Authorize]`.
- Writes stay closed: the CRUD base's `[RequireDeclaredWriteScope]` hard-401s anonymous callers and demands the `*.readwrite` scope, and `RequireScope` independently refuses any non-read requirement for an unauthenticated caller. Empty-trie callers (bare anonymous on the tenant host, Denied-role members) are still rejected by the fallback policy.
- `V4CrudControllerBase.ListDeleted` now carries `[RequireDeclaredWriteScope]`: the recycle bin exists to feed Restore, and owner-deleted records must not become enumerable by the share principal the new class-level read gate admits (review finding).
- New guard: `ShareReadSurfaceReachabilityTests` pins the surface — no `IAuthorizeData` may reappear (that is exactly how this bug shipped), every non-write-gated action (POST computes included) must require its category's read scope, a write-gated read verb must be declared member-only, and every required scope must be in `PublicShareScopes`. Mutation-verified both ways (re-adding `[Authorize]` and stripping a POST gate each fail).
- Frontend: `getPatientTimeZone` falls back to UTC on 401/403; the sleep and IDP reports (never-shareable categories: `sleep.read`, therapy) are hidden from anonymous viewers via a `memberOnly` flag in the report catalog instead of rendering guaranteed errors.

## Deliberate behavior changes to note

- **Members:** a role lacking a category read scope now gets 403 on those endpoints where bare `[Authorize]` used to admit any member — the first slice of the documented Denied-role gap in `AuthorizationConfiguration`. All seed roles except Denied hold the scopes these reports need; a custom role without e.g. `treatments.read` loses the treatments report (which its role editor already claimed it couldn't see).
- **Reading the recycle bin now requires the write scope** (`GET {route}/deleted`), consistent with it being an editing surface.

## Follow-ups (not in this PR)

- Scope-aware navigation: the reports menu still shows categories the viewer wasn't granted (a glucose-only share 403s on the reports overview; a Viewer member 403s on treatments/devices reports). Needs the share's granted categories exposed to the frontend, e.g. via the status endpoint.
- The statistics compute POSTs (`[RequestSizeLimit(100MB)]`) are now reachable by anyone holding a reports-granted share link — pure compute, nothing persisted, but worth considering in any future rate-limiting pass.

## Testing

- All 551 authorization unit tests pass (547 existing + 4 new); full API unit suite has only the two pre-existing failures also present on `main` (decomposer mock, memory benchmark).
- `pnpm run check` stays at the pre-existing 64-error baseline, none in touched files.

https://claude.ai/code/session_01KBFAadoeoM79PCH6CsE665

### DIFF
--- a/src/API/Nocturne.API/Controllers/V4/Analytics/SensorIntegrityController.cs
+++ b/src/API/Nocturne.API/Controllers/V4/Analytics/SensorIntegrityController.cs
@@ -1,5 +1,6 @@
-using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
+using Nocturne.API.Attributes;
+using Nocturne.Core.Models.Authorization;
 using Nocturne.Core.Contracts.Analytics;
 using Nocturne.Core.Models.Analytics;
 using OpenApi.Remote.Attributes;
@@ -20,7 +21,7 @@ namespace Nocturne.API.Controllers.V4.Analytics;
 [Tags("Analytics")]
 [Route("api/v4/sensor-integrity")]
 [Produces("application/json")]
-[Authorize]
+[RequireScope(OAuthScopes.ReportsRead)]
 public class SensorIntegrityController : ControllerBase
 {
     private readonly ISensorIntegrityService _sensorIntegrityService;
--- a/src/API/Nocturne.API/Controllers/V4/Analytics/StatisticsController.cs
+++ b/src/API/Nocturne.API/Controllers/V4/Analytics/StatisticsController.cs
@@ -1,5 +1,6 @@
-using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
+using Nocturne.API.Attributes;
+using Nocturne.Core.Models.Authorization;
 using Nocturne.Core.Constants;
 using Nocturne.Core.Contracts.Analytics;
 using Nocturne.Core.Contracts.Glucose;
@@ -41,7 +42,6 @@ namespace Nocturne.API.Controllers.V4.Analytics;
 [Tags("Analytics")]
 [Route("api/v4/[controller]")]
 [Produces("application/json")]
-[Authorize]
 public class StatisticsController : ControllerBase
 {
     private readonly IStatisticsService _statisticsService;
@@ -114,6 +114,7 @@ public class StatisticsController : ControllerBase
     /// <param name="values">Array of glucose values in mg/dL</param>
     /// <returns>Basic glucose statistics including mean, median, percentiles, etc.</returns>
     [HttpPost("basic-stats")]
+    [RequireScope(OAuthScopes.ReportsRead)]
     public ActionResult<BasicGlucoseStats> CalculateBasicStats([FromBody] double[] values)
     {
         try
@@ -133,6 +134,7 @@ public class StatisticsController : ControllerBase
     /// <param name="request">Request containing glucose values and entries</param>
     /// <returns>Comprehensive glycemic variability metrics</returns>
     [HttpPost("glycemic-variability")]
+    [RequireScope(OAuthScopes.ReportsRead)]
     public ActionResult<GlycemicVariability> CalculateGlycemicVariability(
         [FromBody] GlycemicVariabilityRequest request
     )
@@ -157,6 +159,7 @@ public class StatisticsController : ControllerBase
     /// <param name="request">Request containing entries and optional thresholds</param>
     /// <returns>Time in range metrics including percentages, durations, and episodes</returns>
     [HttpPost("time-in-range")]
+    [RequireScope(OAuthScopes.ReportsRead)]
     [RemoteQuery]
     [RequestSizeLimit(100 * 1024 * 1024)] // 100 MB
     public ActionResult<TimeInRangeMetrics> CalculateTimeInRange(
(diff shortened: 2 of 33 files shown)


bastiaanv:
I've installed tag: `pr-649-9dbd316` on my machine and AGP & Glucose distribution now works!

The following report pages still do not work:
- Executive Summary Report (empty page -> Also empty when using normal login)
- Overview (only shows skeleton loaders)
- Data Quality (shows 401 error)
- Month to month (show error: Could not load tracker data)
- Treatments Report (empty page  -> Also empty when using normal login)
- Insulin Delivery Report (empty page  -> Also empty when using normal login)
- Site Change Impact Report (empty page  -> Also empty when using normal login)
- Idp Report (empty page  -> Also empty when using normal login)

ryceg:
Thanks @bastiaanv — that list was exactly what was needed. I rebased onto `main` (which now carries #650, #651, #661, #663, #664, #666, #668 — none of them touch this) and reproduced your report against a locally seeded tenant (6166 entries, 1522 treatments, 21 tracker instances) with a share link granted all seven categories.

**Two of the eight were real share bugs, and are now fixed** (`c1325d5`):

| Page | Cause | Now |
|---|---|---|
| Data Quality (401) | `CompressionLowController` carried a class-level `[Authorize]` | `RequireScope(glucose.read)`; 401 → 200 |
| Month to Month ("Could not load tracker data") | `/reports/month-to-month` redirects to `/calendar`, which calls `trackers/instances/history` — the one tracker read still carrying `[Authorize]` while its `active` and `upcoming` siblings are `[AllowAnonymous]` | `[AllowAnonymous]` like its siblings; 401 → 200 |

Member reads are unchanged (17 history rows before and after) and the share's writes still 401.

The calendar also had a second bug worth its own mention: **any** tracker query failure replaced the entire month grid with "Could not load tracker data / Calendar view is still available" — the copy promised the calendar and then removed it. Trackers only overlay the grid, which is served by a separate punch-card query, so it now renders with a notice instead.

Root cause of both misses: `ShareReadSurfaceReachabilityTests` pins a hand-maintained list of controllers, so a report page calling a controller nobody listed passes the guard by discovering nothing. Both surfaces are now pinned, plus a guard that a listed controller which stops resolving (rename/move) fails rather than silently dropping out. All three mutation-verified.

**The other six I could not reproduce, and I don't think they're share bugs.** You noted five of them are "also empty when using normal login" — Executive Summary, Treatments, Insulin Delivery, Site Change Impact and IDP — which rules out the share principal. I called their endpoints as a member on the seeded tenant and they all return populated data (`insulin-delivery-stats`, `hourly-insulin-delivery`, `daily-basal-bolus-ratios`, `range-analytics` are all rich), so the API side is healthy on synthetic data. Overview (`/reports`) runs the same `getReportsData` query as AGP, which you say works. These want their own issue against your data rather than being folded in here — if you can share which date range was selected and anything from the browser console/network tab on one of them, that would pin it down.

**One thing worth checking on your side first:** a share link without *full history* is clamped by RLS to the last 24 hours. On the seeded tenant that turned 6166 glucose rows into 286. Every report defaults to a 7–30 day window, so on a default share they render one day of data in a 14-day frame — which can read as "empty". Was full history enabled on the link you tested?

Remaining known gap, not user-visible today: `SummaryController.GetSummary` (`/api/v4/summary`) also 401s a share, but nothing in the app calls it — flagging it rather than widening this PR.

Issue #660, opened 2026-08-07
Two client papercuts: silent widget auth failure, deprecated execCommand in CMS drag handle


ryceg:
Two small, independent client-side fixes, grouped because neither warrants its own issue.

### 1. Windows 11 widget swallows a device-flow failure

`src/Widgets/Nocturne.Widget.Windows11/WidgetProvider.cs`, in the authentication handler:

```csharp
var result = await _oauthService.InitiateDeviceAuthorizationAsync(apiUrl);

if (!result.Success)
{
    _logger.LogWarning("Failed to initiate device authorization: {Error}", result.Error);
    return;
}
```

On failure it logs and returns. The widget is left sitting on the connect card with no state change and no message — from the user's side, pressing the button did nothing. A wrong API URL or an unreachable server is indistinguishable from an unresponsive widget.

`result.Error` is already in hand; it needs surfacing to a widget state the connect card can render.

### 2. CMS drag handle uses deprecated `document.execCommand('copy')`

`src/Web/packages/cms/src/lib/components/edra/shadcn/drag-handle-extended.svelte`, in `handleCopyToClipboard`:

```ts
editor.chain().setMeta('hideDragHandle', true).setNodeSelection(currentNodePos).run();
document.execCommand('copy');
```

`document.execCommand` is deprecated. Replace with `navigator.clipboard.writeText(...)`, taking the selected node's text from the editor rather than relying on the document selection. Note this is async, so the handler needs to await it.

These components originate from edra but are vendored into our tree and re-exported from `@nocturne/cms` (`package.json` → `./editor/*`), so the fix belongs here.

Issue #659, opened 2026-08-07
Bot ack_alert action runs against the unscoped ambient API client


ryceg:
`registerAlertCommands` in `src/Web/packages/bot/src/commands/alerts.ts` handles the `ack_alert` action by calling `getApi()` directly:

```ts
bot.onAction("ack_alert", async (event) => {
  const api = getApi();
  await api.alerts.acknowledge({ acknowledgedBy: event.user.fullName ?? "Unknown" });
  ...
```

Slash commands go through `requireLink` to resolve the invoking chat user to a tenant. Actions don't, because `ActionEvent` has a different shape from `SlashCommandEvent` — no `channel`, `text` or `command` — so it can't be passed to `requireLink` as-is.

The result is that tapping "acknowledge" on an alert notification calls the ambient, unscoped API client. With no tenant in scope it fails, and the user gets "Failed to acknowledge. Please try again." — which will keep failing however many times they retry.

For a chat user linked to more than one tenant, the ambient client is also the wrong thing to be reaching for on an acknowledgement: it isn't derived from who tapped the button.

`mute_30` and `/alerts` in the same file are stubs that post "not yet available" / "coming soon", so they're honest about it; `ack_alert` is the one that looks functional and isn't.

### Suggested resolution

Add a `requireLink` variant that takes the fields an `ActionEvent` does carry (the platform user id is the part that matters) and route `ack_alert` through it, so the acknowledgement lands on the tenant belonging to whoever tapped it.

Issue #657, opened 2026-08-07
Glooko connector keeps per-sync state on instance fields, unsafe for concurrent syncs


ryceg:
`GlookoConnectorService` (`src/Connectors/Nocturne.Connectors.Glooko/Services/GlookoConnectorService.cs`) holds an entire sync's working state on instance fields, populated by `InitializeMappers` during `PerformSyncInternalAsync`:

```csharp
private string? _sessionCookie;
private GlookoUserData? _userData;
private GlookoConnectorConfiguration? _syncConfig;
private GlookoTimeMapper? _timeMapper;
private GlookoSensorGlucoseMapper? _sensorGlucoseMapper;
private GlookoV4TreatmentMapper? _v4TreatmentMapper;
private GlookoStateSpanMapper? _stateSpanMapper;
private GlookoTempBasalMapper? _tempBasalMapper;
private GlookoSystemEventMapper? _systemEventMapper;
private GlookoProfileMapper? _profileMapper;
```

If one instance is ever reached by two overlapping syncs, the second `InitializeMappers` overwrites the first sync's mappers and `_sessionCookie` mid-flight. `_sessionCookie` and `_userData` are the sharp end: a sync could continue against another tenant's Glooko session, and `_syncConfig` drives the time mapping, so records could be written with the wrong timezone timeline.

Whether this is reachable today depends on the connector's DI lifetime and whether the scheduler can overlap runs for one tenant or share an instance across tenants — that should be confirmed as the first step. Either way the class shouldn't be one lifetime change away from cross-tenant data mixing.

### Suggested resolution

Thread the state through explicitly instead of storing it: build a per-sync context object in `PerformSyncInternalAsync` and pass it to the helpers that currently read the fields. That makes the safety property structural rather than dependent on registration.

Issue #656, opened 2026-08-07
Connector catch-up: GetLatestActivityTimestampAsync ignores its source argument


ryceg:
`MetadataPublisher.GetLatestActivityTimestampAsync(string source, ...)` in `src/API/Nocturne.API/Services/ConnectorPublishing/MetadataPublisher.cs` takes a `source` parameter and never uses it. It calls `_activityService.GetActivitiesAsync(count: 1, skip: 0, ...)` and returns the timestamp of the globally newest activity record, whichever connector wrote it.

Activities are stored decomposed across StateSpans, HeartRate and StepCount, and `IActivityService.GetActivitiesAsync` merges and orders them newest-first — so "the latest record" is genuinely global, not per-source.

The consequence for a tenant running more than one activity-producing connector: connector A syncs and writes a recent record, then connector B asks for *its* low-water mark and is handed A's timestamp. B concludes it is already caught up and skips the window it actually still needs, silently leaving a gap.

A single-connector tenant is unaffected, which is why this hasn't surfaced.

`ITreatmentPublisher.GetLatestTreatmentTimestampAsync` has the same shape and may want the same treatment — worth checking as part of this.

### Suggested resolution

Filter by `source` in the query. That needs source-aware reads on the decomposed activity tables, so it's more than a one-line change.

A test with two sources — write a recent record under source A, then assert source B's timestamp is unaffected — pins it. Flip the filter off and the test must fail.

PR #661, opened 2026-08-07
chore: lift TODO markers into issues, and fix meal matching's legacy treatment ids

### LINKED ISSUE #658: Meal-match accept writes a legacy Treatment id into TreatmentFood.CarbIntakeId
`MealMatchingService.AcceptMatchAsync(Guid foodEntryId, Guid treatmentId, ...)` in `src/API/Nocturne.API/Services/Treatments/MealMatchingService.cs` assigns its `treatmentId` argument straight to `TreatmentFood.CarbIntakeId`. Whether that is correct depends entirely on the caller, and the two callers disagree:

- **Glooko** (`GlookoConnectorService.cs`, the food-attribution loop) passes `carbIntake.Id` — a real CarbIntake id. Correct.
- **The suggestion path** passes a **legacy Treatment id**. `FindBestMatch` scores over `IReadOnlyList<Treatment>` and returns a `Treatment`; that treatment goes to `CreateMatchNotificationAsync`, surfaces to the client, and comes back as `AcceptMatchRequest.TreatmentId` at `MealMatchingController.AcceptMatch` — which forwards it unchanged.

So accepting a suggested match writes a `treatment_foods` row whose `carb_intake_id` points at a Treatment, not a CarbIntake. The parameter name (`treatmentId`) reflects the pre-migration model and is now actively misleading at one of the two call sites.

Downstream, `NutritionController` reads `treatment_foods` keyed by carb intake and the carb breakdown is a COB input, so a mismatched key means the breakdown either doesn't resolve or resolves against the wrong record.

### Suggested resolution

Move the whole matching path onto CarbIntake: have the suggestion side score over CarbIntake records rather than `Treatment`, and rename the parameter to `carbIntakeId` so the contract states what it means. `IMealMatchingService.AcceptMatchAsync` in `src/Core/Nocturne.Core.Contracts/Treatments/IMealMatchingService.cs` needs the same rename.

A test that accepts a suggested match and then reads the carb breakdown back through the nutrition path would catch the mismatch; asserting only that a `treatment_foods` row was written will not.

ryceg:
Two commits. The first sweeps every TODO/FIXME marker out of `src/` into the issue tracker. The second fixes one of the things the sweep turned up — read that one on its own, it's a behaviour change.

## 1. `chore:` lift TODO markers into tracked issues

Comments only, no behaviour change.

| Issue | Site |
|---|---|
| #652 | `WebhookSettingsController`, `UISettingsController` — webhook settings and alarm profile writes are silent no-ops |
| #653 | `TrackersController` — `RoleRestricted` hides a tracker from its own owner |
| #654 | `AlertReplayService` — replay should evaluate through `IAlertEvaluationEngine` |
| #655 | `ChatIdentityPendingLinkService` — expired pending-link tokens are never swept |
| #656 | `MetadataPublisher` — `GetLatestActivityTimestampAsync` ignores its `source` argument |
| #657 | `GlookoConnectorService` — per-sync state on instance fields |
| #658 | `MealMatchingService` — accept writes a legacy Treatment id into `TreatmentFood.CarbIntakeId` |
| #659 | `bot/commands/alerts.ts` — `ack_alert` runs against the unscoped ambient API client |
| #660 | `WidgetProvider`, CMS `drag-handle-extended` — silent widget auth failure, deprecated `execCommand` |

No issue numbers in the code — a `#NNN` in a comment is just a TODO wearing a hat. Where behaviour is surprising on its own terms the marker becomes a plain comment stating what the code does; where an adjacent log line or XML doc already said it, the marker is deleted.

Deleted without an issue: `LoopService`'s commented-out IDisposable-template finalizer, and `NocturneDbContext`'s `TODO(Task 1.5)` note, which was **stale** — `CreateLinkAsync` already resolves label collisions via `ResolveUniqueLabel`.

## 2. `fix:` match on carb intakes instead of projected treatments

Closes #658. Verifying that marker turned up a second, larger defect.

Matching read candidates through `ITreatmentStore`, which projects legacy `Treatment` records out of the decomposed v4 tables:

- **`Treatment.DbId` is never assigned anywhere in the codebase.** So `if (treatment.DbId == null) continue;` in `GetSuggestionsAsync` skipped every candidate — `GET /api/v4/meal-matching/suggestions` has been returning `[]` unconditionally, logging a warning per skipped row.
- **On the live path, the id was a bolus id.** The notification metadata carried `treatment.Id`, and per `TreatmentReadService` a CarbIntake paired into a meal bolus is projected with the *bolus's* id. Accepting such a suggestion wrote that bolus id into `treatment_foods.carb_intake_id`. Only unpaired carb corrections happened to carry a real carb intake id — which is why #658's marker described it as a naming problem.

`MealMatchingService` now reads from `ICarbIntakeRepository`; `IsMatch`/`CalculateMatchScore`/`FindBestMatch` take `CarbIntake`. Renamed through the chain so names describe what they hold — `SuggestedMealMatchResult`, `SuggestedMealMatch`, `AcceptMatchRequest.CarbIntakeId`, the notification metadata keys, and the frontend.

Dead code removed: `ProcessNewTreatmentAsync` (no callers) and `Treatment.DbId` (never assigned; this was its last reader).

### Please check these three

- **`AcceptMatchRequest` and `SuggestedMealMatch` are v4 wire contracts.** Property names change. Fine for our own UI, breaking for any third-party consumer.
- **Notification metadata keys change.** Suggestion notifications already in the database can't be accepted, and get re-raised on the next connector sync. I chose that over honouring the old key, because those rows hold a bolus id and accepting one writes the bad FK. Anyone with a pending suggestion sees one "Missing required data" until the resync.
- **One user-facing string changed** — "Treatment has {n}g total carbs" → "The logged carb entry has {n}g total carbs", since "treatment" no longer describes the record. Needs a translation sync.

I stopped short of renaming `treatmentId` on the glucose-chart marker components (`IobCobTrack`, `DeviceEventMarkers`): those genuinely carry projected treatment ids for cross-entity lookup via `getEntryByTreatmentId`.

### Verification

- `dotnet build src/API/Nocturne.API/Nocturne.API.csproj` — 0 errors.
- `dotnet test --filter "FullyQualifiedName~MealMatching"` — 5 passed.
- New test `GetSuggestionsAsync_ReturnsTheMatchedCarbIntakeId` pins the id end to end. **Mutation-checked**: replacing `CarbIntakeId: carbIntake.Id` with a fresh GUID fails it, so it isn't passing by discovering nothing.
- `pnpm run check` in `packages/app` — 64 errors, all in files this branch doesn't touch (generated clients, unbuilt `@nocturne/bot`/`@nocturne/bridge`, `RuleBuilderLeafEditor`, `SwimLaneTrack`). Typechecking caught a second consumer I'd missed, `routes/(authenticated)/meals/+page.svelte`, which is fixed here.
- `git grep -nE "TODO|FIXME" -- src` returns nothing.

https://claude.ai/code/session_01CbBp4z56n5WQ43NGV7m6nq

### DIFF
--- a/src/API/Nocturne.API/Controllers/V4/Connectors/WebhookSettingsController.cs
+++ b/src/API/Nocturne.API/Controllers/V4/Connectors/WebhookSettingsController.cs
@@ -34,7 +34,8 @@ public class WebhookSettingsController(
         CancellationToken cancellationToken = default
     )
     {
-        // TODO: Re-implement with new alert engine storage
+        // Not wired to the current alert engine's storage, so this reports disabled
+        // regardless of what the tenant configured.
         return Task.FromResult<ActionResult<WebhookNotificationSettings>>(Ok(
             new WebhookNotificationSettings
             {
@@ -57,7 +58,6 @@ public class WebhookSettingsController(
         CancellationToken cancellationToken = default
     )
     {
-        // TODO: Re-implement with new alert engine storage
         logger.LogWarning("Webhook settings save is a no-op until new alert engine is implemented");
         return Task.FromResult<ActionResult<WebhookNotificationSettings>>(Ok(settings));
     }
--- a/src/API/Nocturne.API/Controllers/V4/Monitoring/TrackersController.cs
+++ b/src/API/Nocturne.API/Controllers/V4/Monitoring/TrackersController.cs
@@ -93,7 +93,8 @@ public class TrackersController : ControllerBase, IWriteScopedController
         if (tracker.Visibility == TrackerVisibility.Private && tracker.UserId == currentUserId)
             return true;
 
-        // TODO: RoleRestricted visibility check
+        // RoleRestricted has no check yet, so it falls through to hidden — including from
+        // the tracker's own owner.
         return false;
     }
 
--- a/src/API/Nocturne.API/Controllers/V4/Profiles/UISettingsController.cs
+++ b/src/API/Nocturne.API/Controllers/V4/Profiles/UISettingsController.cs
@@ -367,7 +367,8 @@ public class UISettingsController : ControllerBase, IWriteScopedController
                 );
             }
 
-            // TODO: Re-implement with new alert engine
+            // The profile is discarded: this is not wired to the current alert engine's
+            // storage.
             return await GetAlarmConfiguration(cancellationToken);
         }
         catch (Exception ex)
@@ -405,8 +406,8 @@ public class UISettingsController : ControllerBase, IWriteScopedController
                 );
             }
 
-            // TODO: Re-implement with new alert engine
-
+            // The profile id is discarded: this is not wired to the current alert engine's
+            // storage.
             return await GetAlarmConfiguration(cancellationToken);
         }
         catch (Exception ex)
--- a/src/API/Nocturne.API/Controllers/V4/Treatments/MealMatchingController.cs
+++ b/src/API/Nocturne.API/Controllers/V4/Treatments/MealMatchingController.cs
@@ -97,9 +97,9 @@ public class MealMatchingController : ControllerBase
             MealName = s.MealName,
             Carbs = s.Carbs,
             ConsumedAt = s.ConsumedAt,
-            TreatmentId = s.TreatmentId,
-            TreatmentCarbs = s.TreatmentCarbs,
-            TreatmentMills = s.TreatmentMills,
+            CarbIntakeId = s.CarbIntakeId,
+            CarbIntakeCarbs = s.CarbIntakeCarbs,
+            CarbIntakeMills = s.CarbIntakeMills,
             MatchScore = s.MatchScore,
         }).ToArray();
 
@@ -132,7 +132,7 @@ public class MealMatchingController : ControllerBase
         {
             await _mealMatchingService.AcceptMatchAsync(
                 request.FoodEntryId,
-                request.TreatmentId,
+                request.CarbIntakeId,
                 request.Carbs,
                 request.TimeOffsetMinutes,
                 HttpContext.RequestAborted);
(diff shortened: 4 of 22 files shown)


ryceg:
Review pass done; third commit pushed with the fixes.

**Fixed — candidate fetch direction (regression this branch introduced).** `GetCarbIntakesInWindowAsync` passed `descending: false`, so a window holding more than `CandidateLimit` carb intakes kept the *oldest* 1000 and dropped the newest. The old projection path read every v4 repo newest-first and took the most recent `limit`, so this was a behaviour change I made by accident rather than an inherited quirk. It matters because `GetSuggestionsAsync` fetches one candidate set for the caller's entire range and `/meals` passes a user-chosen range straight through — a wide range would return suggestions only for the far end, while the pending MyFitnessPal entries are always the recent ones. Now `descending: true`.

**Fixed — accept writes without checking the carb intake exists.** `treatment_foods.carb_intake_id` has no foreign key, so posting an arbitrary or stale Guid to `POST /meal-matching/accept` persisted a row no breakdown could resolve *and* flipped the food entry to `Matched`, so it never resurfaced. Worth noting this was unreachable before — accept couldn't be driven while suggestions were always empty — so it goes live with this branch rather than being pre-existing. `AcceptMatchAsync` now returns early if the carb intake isn't found.

Both new tests are mutation-checked: flipping `descending` back to `false` and deleting the existence check each fail their test.

**Not fixed — stale notification metadata.** Still no fallback to the old `treatmentId` key, so suggestion notifications sitting in `in_app_notifications` at deploy time go inert until the next connector sync re-raises them. The review's own detail is what settled it for me: those rows carry `treatment.Id`, which for a paired meal bolus is the *bolus* id. A fallback would make exactly the broken ones acceptable again, and each acceptance writes the orphan row the second fix above now rejects. I'd rather they fail visibly and get re-raised correctly. Happy to add a one-off archive of active `meal_matching.suggested_match` rows on deploy if you'd prefer the dead-notification window closed.

Also worth recording from the review, since it contradicts a docs claim: `ICarbIntakeRepository.GetAsync`'s XML doc says `to` is exclusive, but `CarbIntakeRepository` applies `e.Timestamp <= to.Value`. The bound is inclusive, matching the old `date[$lte]`, so the window is unchanged — but the doc is wrong and will mislead the next caller.

PR #871, opened 2026-08-19
fix: update actograms

### LINKED ISSUE #761: Actogram reports
Today, I've been working on getting my Apple Health data into Nocturne and got the following reports:

Steps:

Image

Heart rate:

Image

I was wondering if the following could be changed:
- The Heart rate actogram is in a reversed order compared to the steps. Could we have on the heart rate newest date at the top?
- Both actogram's show the glucose in mg/dl, could this be shown based on the unit in Appearance
- Both actogram's show the datetime in 12h format, could this also be shown based on the time format in Appearance

bastiaanv:
Changes:
- fix: All actograms are showing the correct glucose & time units
- fix: swap heart rate time ordering (newest to oldest)

Closes #761


ryceg:
Thanks Bastiaan!

PR #997, opened 2026-08-27
fix(aspire): wire RESEND_* env vars into the web resource

BlackSoul697:
## Summary

The web app's bot init (`src/Web/packages/app/src/lib/server/bot/index.ts`) enables the Resend email-alert adapter from `RESEND_API_KEY` / `RESEND_FROM_ADDRESS`, and the docs (`docs/alerts/email.svx`, "Option B") document `RESEND_API_KEY`, `RESEND_FROM_ADDRESS`, `RESEND_FROM_NAME` and `RESEND_WEBHOOK_SECRET` as environment configuration for deployments. But neither the Aspire web resource (`ConfigureWebEnvironment`) nor the committed compose bundle's `.env.example` carried those variables, so the documented configuration path was dead in every generated deployment — only the admin-UI path (platform settings in the DB) worked.

## Changes

- `src/Aspire/Nocturne.Aspire.Host/Program.cs`: add the four optional parameters (`resend-api-key` and `resend-webhook-secret` as secrets) and wire them into `ConfigureWebEnvironment` alongside the other chat-platform credentials.
- `deploy/docker-compose/.env.example`: document the four optional `RESEND_*` variables.

## Context

- Consumed by the bot init: `src/Web/packages/app/src/lib/server/bot/index.ts` and `src/Web/packages/bot/src/bot.ts`.
- Documented in `src/Web/packages/portal/src/content/docs/alerts/email.svx`.


ryceg:

---

**Requesting changes** — the `Program.cs` half is correct, but the other half is a no-op that will be automatically reverted, so the bug this PR describes survives the fix.

**1. The documented path is still dead after this merges.** Neither compose file uses `env_file:`, so a container only receives what's in its own `environment:` map. `deploy/docker-compose/docker-compose.yaml:78` carries `WHATSAPP_ACCESS_TOKEN: "${WHATSAPP_ACCESS_TOKEN}"`; there is no `RESEND_*` equivalent and this PR doesn't add one. An operator who sets `RESEND_API_KEY` in `.env` gets nothing in the container — exactly the failure described in the summary.

**2. `.env.example` is generated, not authored.** `scripts/publish-release.cs:102` writes it from `GenerateEnvExample(...)`, and line 109 writes *the same string* to `deploy/portainer/.env.example`. So the hand-edit:

- diverges two files the generator guarantees are byte-identical (only one was edited);
- gets silently clobbered on the next release publish;
- can't survive in the shape it's written. The generator emits a flat `# -- Optional ---` section of bare `# NAME=` lines (`publish-release.cs:241-244`). The `# -- Email alerts (Resend) --` header, the prose comment, and the `alerts@your-instance.com` example value have no generator equivalent and will all be deleted.

**3. Three further generated artifacts go stale:** `deploy/portainer/docker-compose.yaml`, `deploy/portainer/.env.example`, and `deploy/portainer/templates.json` (built from the same `groups`, line 110).

### What to do

Keep the `Program.cs` change exactly as-is, drop the `.env.example` hunk entirely, run `scripts/publish-release.cs`, and commit all five regenerated files. If you want `RESEND_FROM_ADDRESS` to carry an example value or a section header, that has to come from the generator (or `WithPublishMetadata`), not a hand-edit.

### Smaller points

- Rationale is duplicated at two sites (the parameter declarations and again at `.WithEnvironment`). House convention is one site.
- The new parameters sit *outside* the existing `// Chat platform credentials. All optional…` block at `Program.cs:252-256`, which already explains precisely this. Fold them in.
- `RESEND_WEBHOOK_SECRET` has no reader anywhere in this repo — `bot.ts:120` asserts the adapter auto-loads it from env. Please confirm that's true, otherwise we're wiring and documenting a dead variable.

The wiring itself checks out: `$env/dynamic/private` (`bot/index.ts:3`) means runtime injection works, and the names match `index.ts:85` / `bot.ts:118-119`.

BlackSoul697:
Thanks for the detailed review — all three points addressed. I ran `scripts/publish-release.cs` and committed the regenerated files (`df24d400`).

**1. Documented path now live.** `RESEND_API_KEY`, `RESEND_FROM_ADDRESS`, `RESEND_FROM_NAME`, `RESEND_WEBHOOK_SECRET` now appear in the web service's `environment:` map in **both** compose bundles (`deploy/docker-compose/docker-compose.yaml` and `deploy/portainer/docker-compose.yaml`), interpolated as `${RESEND_*}` exactly like the existing `WHATSAPP_*` entries. An operator setting these in `.env` now gets them in the container.

**2. No more hand-edit.** I dropped the authored `.env.example` hunk. The four variables are now added to the generator's `optional` set in `scripts/publish-release.cs`, so `GenerateEnvExample` emits them as flat `# RESEND_*=` lines in the `-- Optional --` section. After the generator ran, `deploy/docker-compose/.env.example` and `deploy/portainer/.env.example` are byte-identical (verified via `cmp`), and the format matches every other optional variable. Any desired example value or header would have to come from the generator / `WithPublishMetadata` as you noted.

**3. Regenerated artifacts committed.** All of `deploy/docker-compose/docker-compose.yaml`, `deploy/docker-compose/.env.example`, `deploy/portainer/docker-compose.yaml`, `deploy/portainer/.env.example` were regenerated and committed. `deploy/portainer/templates.json` came out byte-identical to its prior state and is intentionally not touched.

**Smaller points:**

- **Single rationale site:** Removed the Resend comment block from the parameter declarations and the block at `.WithEnvironment`; the Resend parameters now live inside the existing `// Chat platform credentials. All optional …` block, which already explains the pattern.
- **RESEND_WEBHOOK_SECRET reader confirmed:** `src/Web/packages/bot/src/bot.ts:117-120` calls `createResendAdapter({ fromAddress, fromName })` and, when both `apiKey` and `webhookSecret` are omitted from the options, relies on the adapter auto-loading them from env. Confirmed against `@resend/chat-sdk-adapter@^0.2.2`: its README and the Resend Chat SDK docs state `RESEND_API_KEY` and `RESEND_WEBHOOK_SECRET` are read from the environment when not passed in options. So wiring/documenting them is correct — they have a live reader.

Do note the `docker-compose.yaml` bundles also pick up the current `SCALAR_CUSTOM_CSS` value (source theme drift since the last full regenerate) — that is genuine generator output, not a regression.

Uneo7:
I tested your merge request, it's all good thanks for contributing :) 

(I rebased and fixed the conflicts with main)

Uneo7 (approved):

PR #998, opened 2026-08-27
fix(helm): SQL-escape passwords in the bootstrap Job SET statements

BlackSoul697:
## Summary

`deploy/helm/nocturne/templates/bootstrap-configmap.yaml`'s `run.sh` builds the `SET nocturne.<role>_password` literals with a naive `printf "'%s'"`, so a role password containing a single quote produces broken SQL (an unterminated string literal, e.g. `SET nocturne.web_password TO 'pw'rd';`) and the bootstrap Job fails under `ON_ERROR_STOP=1` — or, worse, the trailing password text is parsed as SQL.

## Changes

- `deploy/helm/nocturne/templates/bootstrap-configmap.yaml`: add an `escape_sql_literal()` helper (doubles single quotes, SQL-standard) and route all three `SET` statements through it.
- `scripts/test-helm-bootstrap-escaping.sh`: regression test that extracts `run.sh` verbatim from the template, runs it against mock `pg_isready`/`psql` binaries, and asserts the apostrophes are doubled.

## Verification

- `bash scripts/test-helm-bootstrap-escaping.sh` passes on the fixed template and fails against the pre-fix template (exit 1), confirming the test is regression-sensitive.

### DIFF
--- a/deploy/helm/nocturne/files/bootstrap-roles.sql
+++ b/deploy/helm/nocturne/files/bootstrap-roles.sql
@@ -18,10 +18,15 @@
 
 \set ON_ERROR_STOP on
 
--- Passwords come in via PostgreSQL session custom-GUCs set by run.sh
--- before this file is sourced. psql's `:'var'` substitution doesn't reach
--- inside dollar-quoted DO blocks (psql 14+), so we read from session
--- settings instead.
+-- Passwords arrive via psql -v variables and are placed into session
+-- custom-GUCs here, OUTSIDE the dollar-quoted DO block (psql :'var'
+-- substitution doesn't reach inside DO blocks, psql 14+). This matches the
+-- compose bundle's `psql -v ... :'var'` pattern: psql quotes the values
+-- safely regardless of embedded quotes, dollar signs or backslashes.
+SELECT set_config('nocturne.migrator_password', :'migrator_password', false);
+SELECT set_config('nocturne.app_password',      :'app_password',      false);
+SELECT set_config('nocturne.web_password',      :'web_password',      false);
+
 DO $$
 DECLARE
     migrator_password text := current_setting('nocturne.migrator_password');
--- a/deploy/helm/nocturne/templates/bootstrap-configmap.yaml
+++ b/deploy/helm/nocturne/templates/bootstrap-configmap.yaml
@@ -40,11 +40,16 @@ data:
     # psql's `:'var'` substitution doesn't reach inside dollar-quoted DO
     # blocks (psql 14+), so we prepend SET statements that the DO block
     # then reads via current_setting('nocturne.<role>_password').
+    # The literal is escaped SQL-style (single quotes doubled) so passwords
+    # containing single quotes cannot break the statement.
+    escape_sql_literal() {
+      printf "%s" "$1" | sed "s/'/''/g"
+    }
     TMPSQL=$(mktemp) && trap 'rm -f "$TMPSQL"' EXIT
     {
-      printf "SET nocturne.migrator_password TO %s;\n" "$(printf "'%s'" "$MIGRATOR_PASSWORD")"
-      printf "SET nocturne.app_password TO %s;\n"      "$(printf "'%s'" "$APP_PASSWORD")"
-      printf "SET nocturne.web_password TO %s;\n"      "$(printf "'%s'" "$WEB_PASSWORD")"
+      printf "SET nocturne.migrator_password TO '%s';\n" "$(escape_sql_literal "$MIGRATOR_PASSWORD")"
+      printf "SET nocturne.app_password TO '%s';\n"      "$(escape_sql_literal "$APP_PASSWORD")"
+      printf "SET nocturne.web_password TO '%s';\n"      "$(escape_sql_literal "$WEB_PASSWORD")"
       cat /scripts/bootstrap-roles.sql
     } > "$TMPSQL"
     exec psql -v ON_ERROR_STOP=1 -f "$TMPSQL"
--- a/deploy/helm/nocturne/templates/bootstrap-configmap.yaml
+++ b/deploy/helm/nocturne/templates/bootstrap-configmap.yaml
@@ -36,21 +36,17 @@ data:
 
     echo "Running Nocturne role bootstrap against ${PGHOST}:${PGPORT}/${PGDATABASE} as ${PGUSER}"
 
-    # Pass passwords into the SQL via PostgreSQL session custom-GUCs.
-    # psql's `:'var'` substitution doesn't reach inside dollar-quoted DO
-    # blocks (psql 14+), so we prepend SET statements that the DO block
-    # then reads via current_setting('nocturne.<role>_password').
-    # The literal is escaped SQL-style (single quotes doubled) so passwords
-    # containing single quotes cannot break the statement.
-    escape_sql_literal() {
-      printf "%s" "$1" | sed "s/'/''/g"
-    }
-    TMPSQL=$(mktemp) && trap 'rm -f "$TMPSQL"' EXIT
-    {
-      printf "SET nocturne.migrator_password TO '%s';\n" "$(escape_sql_literal "$MIGRATOR_PASSWORD")"
-      printf "SET nocturne.app_password TO '%s';\n"      "$(escape_sql_literal "$APP_PASSWORD")"
-      printf "SET nocturne.web_password TO '%s';\n"      "$(escape_sql_literal "$WEB_PASSWORD")"
-      cat /scripts/bootstrap-roles.sql
-    } > "$TMPSQL"
-    exec psql -v ON_ERROR_STOP=1 -f "$TMPSQL"
+    # Pass passwords to psql via -v variables (never shell-interpolated into
+    # the SQL text), matching the compose bundle's bootstrap pattern. psql's
+    # :'var' quoting handles values containing quotes, dollar signs or
+    # backslashes safely; files/bootstrap-roles.sql places them into session
+    # custom-GUCs via set_config(...) OUTSIDE the dollar-quoted DO block
+    # (psql :'var' substitution doesn't reach inside DO blocks). No temp
+    # files: passwords never touch disk, and the `exec` below replaces this
+    # shell so no EXIT trap would fire anyway.
+    exec psql -v ON_ERROR_STOP=1 \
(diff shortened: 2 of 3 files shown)


ryceg:

---

**Requesting changes** — the bug is real and the fix works, but it reinvents something we already solved one directory over, and the regression test can't run in CI.

**1. Use the pattern the compose bundle already uses.** `deploy/docker-compose/docker-compose.yaml:279-287` performs the identical role bootstrap via `psql --set web_password="$..."` with `:'web_password'` in the SQL — and its comment at lines 268-269 says outright *"so that values containing quotes or dollar signs are handled safely by psql's `:'var'` quoting."* We should not have two escaping strategies for the same three passwords.

The Helm side diverged only because `:'var'` can't reach inside `DO $$ … $$`. That's solved by moving the substitution outside the block:

```sql
SELECT set_config('nocturne.migrator_password', :'migrator_password', false);
SELECT set_config('nocturne.app_password',      :'app_password',      false);
SELECT set_config('nocturne.web_password',      :'web_password',      false);
```

with `run.sh` reducing to a single `exec psql -v ON_ERROR_STOP=1 -v migrator_password="$MIGRATOR_PASSWORD" … -f /scripts/bootstrap-roles.sql`. That deletes `escape_sql_literal`, the `mktemp`, the `trap`, the temp file and the `cat` concat, hands escaping to psql, and restores real line numbers in psql errors. It also makes the `files/bootstrap-roles.sql` header comment (lines 4-6) true — it already claims the file reads passwords via `-v`, which it currently doesn't.

**2. The `trap` on the lines you're editing is already dead.** `exec psql …` replaces the shell, so `trap 'rm -f "$TMPSQL"' EXIT` never fires and the temp file holding all three plaintext passwords is never removed. Container-ephemeral so the blast radius is small, but it's a live defect in the block being touched. The `-v` approach never writes them to disk at all.

**3. Nothing runs the new test.** `helm-lint.yml` already exists and already carries a bespoke *"Bootstrap SQL drift check"* step — that's the home for this. As shipped, `scripts/test-helm-bootstrap-escaping.sh` is invoked by no workflow, so the regression is unguarded.

**4. The test writes outside the repo and never cleans up.** `sudo mkdir -p /scripts` + `sudo cp` leave `/scripts/bootstrap-roles.sql` on the host permanently — the `trap` only removes `$work`. The sudo requirement also guarantees most contributors skip it locally. Changing run.sh to read `${SCRIPTS_DIR:-/scripts}/bootstrap-roles.sql` removes the sudo, the host writes and the fragility together.

**5. Coverage gap:** only `'` is exercised. Please add a password with a trailing newline — `$( )` strips it, so we'd silently set a *different* password with no error — plus backslash and `$` cases.

### Out of scope but worth a follow-up issue

`docs/postgres/bootstrap-roles.sql:56-58` has the same hazard class via hand-substituted `'REPLACE_ME_…'` literals, with no warning to the operator.

For the record, the scoping is right: every `CREATE`/`ALTER ROLE` in `files/bootstrap-roles.sql` already goes through `format(…, %L)`, so the `SET` statement genuinely was the only injection point.

BlackSoul697:
All five review points are addressed in ec3a6638f (pushed to this PR branch, no force-push):

1. **Compose pattern adopted** — run.sh is now a single `exec psql -v ON_ERROR_STOP=1 -v migrator_password="$MIGRATOR_PASSWORD" -v app_password=... -v web_password=... -f "${SCRIPTS_DIR:-/scripts}/bootstrap-roles.sql"`, and files/bootstrap-roles.sql sets the session custom-GUCs via `SELECT set_config('nocturne.<role>_password', :'<role>_password', false);` **outside** the DO block (the DO block reads them via current_setting). escape_sql_literal is gone; escaping is psql's `:'var'` quoting, identical to deploy/docker-compose.
2. **Temp file / dead trap eliminated** — no mktemp, no trap, no plaintext SQL on disk at all; the dead `trap ... EXIT` before `exec` is removed with the file it guarded.
3. **CI wiring** — the regression test step was added to helm-lint.yml next to the drift check. Caveat: my token lacks the `workflow` scope, so that one file could not be pushed from here (the rest of the branch is up). The helm-lint.yml change is verified locally and ready — happy to push it once scope is available, or it can land from a maintainer side.
4. **No sudo / no host writes** — run.sh reads `${SCRIPTS_DIR:-/scripts}/bootstrap-roles.sql`; the test uses a temp SCRIPTS_DIR.
5. **Coverage extended** — passwords with `'`, `$`, `\\` and an embedded newline are asserted byte-for-byte in the psql ``-v`` argv; the test also asserts no plaintext ever appears in the SQL body and that no temp-file/mktemp/trap pattern regresses.

Verification evidence:
- `bash scripts/test-helm-bootstrap-escaping.sh` → `OK: bootstrap password handling regression test passed (psql -v, no temp files, SCRIPTS_DIR honoured).`
- `bash -n` on run.sh and the test script: clean.
- CI `lint-and-template` (helm lint + template + kubeconform + drift check): **pass** on this push.
- Out-of-scope note acknowledged: docs/postgres/bootstrap-roles.sql has the same hand-substitution hazard — agree it warrants a follow-up issue rather than widening this PR.

ryceg:

---

**Requesting changes — but the important caveat first: the blocker below is my fault, not yours.** You implemented the `set_config` approach exactly as I specified it in my last review, and it turns out that approach has a hole I didn't spot when I wrote it. Sorry for sending you down that path.

Everything else about this revision is right. I reproduced the original bug against a real `postgres:17-alpine` with `MIGRATOR_PASSWORD=mig'ra`:

```
psql: ERROR:  syntax error at or near "ra"
LINE 1: SET nocturne.migrator_password TO 'mig'ra';
EXIT=3
```

and confirmed your branch fixes it — the role authenticates fine afterwards with a password containing `'`, `$` and `\`. Dropping the `mktemp`/`trap` was the right call too, and it's better than I realised: `containerSecurityContext.readOnlyRootFilesystem` is `false` today only because of a "tighten in a later iteration" TODO, so the temp file was a latent breakage waiting for that TODO to be done.

## Blocker: `set_config` echoes the passwords into the Job's pod logs

This is the part I got wrong. `set_config()` **returns the value it sets**, and `run.sh` invokes `psql` with no `-q`, no `-o` and no redirection. Actual container stdout from this branch:

```
Running Nocturne role bootstrap against localhost:5432/nocturne as postgres
  set_config
--------------
 mig'ra$tor\x
(1 row)

 set_config
------------
 ap'p$
(1 row)

 set_config
------------
 web"pw
(1 row)

DO
```

`main` printed `SET / SET / SET / DO`. So as written this trades a crash for plaintext migrator/app/web passwords in `kubectl logs` and in every log shipper in the cluster — and with `restartPolicy: OnFailure` it re-leaks on each retry. `nocturne_migrator` owns the database and the `public` schema, so that's an escalation path and strictly worse than the bug we're fixing.

Fix is one token per line. I verified the GUC still round-trips exactly with a `"`/`$`-bearing value:

```sql
SELECT set_config('nocturne.migrator_password', :'migrator_password', false) \g /dev/null
SELECT set_config('nocturne.app_password',      :'app_password',      false) \g /dev/null
SELECT set_config('nocturne.web_password',      :'web_password',      false) \g /dev/null
```

While you're in there: my last review said this "matches the compose bundle pattern", and that's only half true — worth correcting the comments in `run.sh` and the SQL so the next reader isn't misled. The compose bundle and `docs/postgres/container-init/00-init.sh` use `:'var'` *directly in the DDL*, with no GUC round-trip and therefore nothing to echo. The `set_config` hand-off here exists solely because `:'var'` can't reach inside `DO $$ … $$` and the Helm copy needs the `DO` block for idempotency. That's a legitimate divergence; it just needs saying as one.

## The test can't catch that, by construction

`scripts/test-helm-bootstrap-escaping.sh` runs `run.sh` as `>/dev/null 2>&1` — it discards the exact stream the secrets leak into. And the "no plaintext leak" assertion:

```bash
if grep -qE "mig'ra|ap'p|we.b" "$work/out/sqlbody.txt"; then
```

only greps `sqlbody.txt`, which is a straight `cat` of the chart's own committed `bootstrap-roles.sql` — a file `run.sh` never writes to. That check cannot fail for any change to `run.sh`. Could you capture stdout instead and fail if any password value appears in it? That's the assertion that would have caught my mistake, and it's the one that matters going forward.

Two related things on the test:

- **It isn't wired into CI.** `grep -rn "test-helm-bootstrap" .github/` returns nothing — I see from your commit message that the `helm-lint.yml` wiring was withheld because your token lacks the `workflow` scope. Completely understood, and that's on our side to unblock: say the word and I'll push the workflow wiring myself so you don't have to fight it. It'll also need `scripts/**` adding to `helm-lint.yml`'s `paths:` filter, which is currently `deploy/helm/**` only. As it stands the script would never run and would quietly rot.
- **Name and scope have drifted.** It's `test-helm-bootstrap-escaping.sh` but there's no escaping left in the change, and the header rightly notes that `:'var'` correctness is Postgres' contract and out of scope. So it's really testing argv plumbing — which is worth testing, it just wants a name like `test-helm-bootstrap-password-handling.sh`.

## Smaller points

**`SCRIPTS_DIR` is a test-only seam in shipped config.** `bootstrap-job.yaml` hardcodes `mountPath: /scripts` and `command: ["/bin/sh", "/scripts/run.sh"]`, so `run.sh`'s own path is fixed while the SQL's is now overridable. I'd rather the harness bind-mount or wrap than have the chart carry an override that only exists for the test. If you'd prefer to keep it, that's arguable — but then it should be documented in `values.yaml` as intentional chart API.

**PR title and body are stale.** The title still says "SQL-escape passwords in the bootstrap Job SET statements", but your third commit removed `escape_sql_literal()` and there are no `SET` statements left. The body still documents the deleted helper, and the Verification section describes running the old test against the old template. We squash-merge, so that description becomes the permanent commit message — could you refresh both?

**Nit:** the test's `mktemp -d` is never cleaned up.

## Summary

1. `\g /dev/null` on the three `set_config` calls — the only true blocker
2. Test asserts on stdout and fails if a password value appears there
3. Test wired into CI (tell me if you want me to push the workflow file), or dropped
4. Rename the test file; refresh the PR title and body
5. `SCRIPTS_DIR` moved into the harness, or documented as intentional

All 14 checks are green, but `lint-and-template` only proves the chart renders — it never executes `run.sh`, which is exactly why (2) matters.

Thanks for your patience through two rounds on this, especially given the second round was chasing my own bad advice. The direction is right and everything left is mechanical.

PR #1000, opened 2026-08-27
fix(web): make local report day boundaries DST-safe

BlackSoul697:
## Summary
- computes UTC report boundaries from actual local calendar midnights
- derives the end from the following local midnight instead of assuming 24 hours
- covers midnight transitions and 23/25-hour days

## Problem
`getLocalDayBoundariesUtc` used locale-string parsing and always added 24 hours. In zones with DST transitions, report windows could overlap the next local day or omit part of a 25-hour day.

## Tests
- `pnpm vitest run src/lib/utils/timezone.test.ts`: 7 passed
- sabotage run with the previous implementation: 3 new DST tests failed
- `pnpm run check`: remaining generated-module errors were also present on the upstream baseline

## Scope
Only the timezone helper and its focused tests are changed. Related profile-unit and bridge-CORS findings are intentionally excluded.

### DIFF
--- a/src/Web/packages/app/src/lib/utils/timezone.test.ts
+++ b/src/Web/packages/app/src/lib/utils/timezone.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from "vitest";
-import { localDayStart, localDayEnd } from "./timezone";
+import { localDayStart, localDayEnd, getLocalDayBoundariesUtc } from "./timezone";
 
 describe("localDayStart", () => {
 	it("returns local midnight, not UTC midnight", () => {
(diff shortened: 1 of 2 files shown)


ryceg:

---

**Approving in principle — one fix needed before merge.**

I verified this rather than reading it. I fuzzed the new `localMidnightUtc` against a binary-search ground truth across 945 DST-adjacent days in twelve hostile zones — Santiago, Havana, Tehran, Lord Howe (30-minute DST), Chatham (+12:45/+13:45), Apia (date-line jump), Troll (2-hour jump), St John's, São Paulo, Beirut, Berlin, Sydney: **zero mismatches**. Old-vs-new on the dates in the tests also shows the bug was slightly worse than the summary claims — Berlin `2026-03-29` old `end` was `22:59:59.999Z`, which formats as local date **2026-03-30**, and Santiago `2027-04-04` old `start` was a full hour early. Good change.

**1. The fall-back comment is wrong, and contradicts itself.** In `timezone.test.ts`:

> Chile falls back 2027-04-04 01:00 -> 00:00 (-03 -> -04); the extra hour repeats at the END of Apr 3

Checked against Intl:

```
2027-04-04T02:59Z -> 2027-04-03, 23:59 GMT-3
2027-04-04T03:00Z -> 2027-04-03, 23:00 GMT-4
```

Local goes **24:00 Apr 3 → 23:00 Apr 3**. There is no `01:00 -> 00:00`, and the mechanism as written would put the repeated hour on Apr 4 — contradicting both its own next clause and the assertion below it. The assertion is right; only the comment is wrong. Please fix it: a wrong comment in a DST test is the one place we can least afford one. (The spring-forward comment is accurate.)

**2. Silent fallback.** `valid.length > 0 ? Math.min(...valid) : c1` returns a known-suspect value when neither probe validates, and two-probe convergence is empirical rather than proven. Cheap hardening using the test's own oracle: assert `tzDateStr(start - 1) !== dateStr`, i.e. that it's the *first* instant of the day and not merely *an* instant with the right date. That's what `firstInstantOfDay` checks and what the implementation doesn't.

**3. Test nit:** `firstInstantOfDay` constructs a fresh `Intl.DateTimeFormat` inside a 1440-iteration loop — hoist it. The 60-second grid also assumes minute-aligned transitions, which is fine for every modern zone but silently wrong for LMT-era dates.

**4. Closest relative left behind:** `GlucoseSparkline.svelte:50` uses the same `start + 24 * 60 * 60 * 1000 - 1` for its x-domain, so the axis is an hour off on 23/25-hour days. The Scope note lists other exclusions but not this one — either fold it in or file it.

**5. Noting for whoever touches this file next** (not this PR's problem): the `!timeZone` branch at `timezone.ts:33-36` returns **UTC** day bounds while `localDayStart`/`localDayEnd` in the same file return **browser-local** bounds. Same file, similar names, different meanings.

Fix #1 and I'm happy to merge; #2–#4 are your call.

PR #1045, opened 2026-08-30
fix(connectors): bind ApiSecret in NocturneRemote configuration

DChau10:
The NocturneRemote connector configuration decorated Token with ConnectorPropertyKey.ApiSecret. The schema generator produced the camel-cased property name "apiSecret" for the UI, while the reflection binder looked for a property named "Token" ("token"). This property name mismatch prevented entered API secrets from binding, causing sync to fail with "Not syncing: ApiSecret is required but not configured."

Rename the configuration property to ApiSecret to match the attribute key name, and retain Token as a compatibility alias. Add configuration binding unit tests to guard against future property name and attribute key drift.

### DIFF
--- a/src/API/Nocturne.API/Services/BackgroundServices/NocturneRemoteConnectorBackgroundService.cs
+++ b/src/API/Nocturne.API/Services/BackgroundServices/NocturneRemoteConnectorBackgroundService.cs
@@ -90,7 +90,7 @@ public class NocturneRemoteConnectorBackgroundService : ConnectorBackgroundServi
                 var connection = new HubConnectionBuilder()
                     .WithUrl(hubUrl, options =>
                     {
-                        options.Headers.Add("Authorization", $"Bearer {config.Token}");
+                        options.Headers.Add("Authorization", $"Bearer {config.AccessToken}");
                     })
                     .WithAutomaticReconnect(new InfiniteRetryPolicy())
                     .Build();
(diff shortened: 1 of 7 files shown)


ryceg:
Hi, thanks for the PR, and good catch! The real problem is that the DB binder is the only thing that matches on the C# property name; the schema, env vars, appsettings and the "missing required" message all use the attribute key. Every other connector only works because the two happen to line up. Fix the binder to read the attribute and the whole class of bug goes away :)

Also, "API secret" isn't our term any more; it's an API token now (noc_ prefix), which is what this connector actually sends as a bearer. So right now someone setting up Nocturne Remote gets a field labelled "API Secret" described as a Nightscout API_SECRET, and pastes a noc_ token into it. There's already an unused ConnectorPropertyKey.AccessToken which should probably be used instead.

And I'd drop the Token alias while we're here. It's only a handful of call sites, and the new guard test doesn't cover the alias anyway. Cheers!

DChau10:
@ryceg That makes sense! I'm admittedly a bit out of my element here with this specific stack and am learning as I go, so I appreciate the feedback as I learn more about the stack and implementation.

I've updated the `ConnectorConfigurationBinder.cs` to resolve the attribute name with a fallback to the property name.

On this branch, I've tested a successful connection to my production nocturne instance with data ingested.

PR #1273, opened 2026-09-08
Add configurable Year Overview color ranges

smokkelaar:
## Summary

Year Overview now supports saved color-focus ranges for TDD, bolus, basal, carbohydrates and Time in Range, plus four adjustable average-glucose color boundaries. Preferences use the existing per-user store, cookie hydration and backend endpoint, and are included in the existing shared-appearance projection. Resets persist as an empty color selection.

A single control handles both range types, with exact numeric input, keyboard and pointer support, accessible color-zone descriptions and a printed legend of the active values. Slider updates insert exact bounds into cached sorted steps instead of sorting the whole array on every move.

The requested low-end palette now transitions from 4 to 3 mmol/L (72 to 54 mg/dL), and stays at its endpoint at or below 3: black on light cards and white on dark cards. Printed Year Overview reports use white paper and the black endpoint, including when printed from dark mode. These are presentation boundaries only; glucose targets and Time in Range calculations do not change.

## Validation

- API build, NSwag client, Zod schemas and remote-function generation passed locally.
- 80 color utility, preference-store and SSR tests passed.
- 17 control browser tests and 9 report-page browser tests passed.
- 4 palette browser cases passed, covering light, dark and print for Nocturne, Trio, AAPS and Classic; screenshots were visually inspected.
- 36 preference-model tests and 11 preference/share API tests passed.
- App typecheck still reports 34 existing diagnostics; none is new relative to the reviewed base when compared by file and diagnostic. The six reported feature errors are resolved.

The repository workflows do not run the app Vitest suites. The results above are actual local runs using the standard configurations, with Windows-safe aliases and Svelte deduplication; no CI coverage is claimed.


smokkelaar:
Addressed the review feedback in 75f3e24: localStorage reads and writes are now guarded by the SvelteKit browser flag and storage availability, while blocked storage still degrades to the existing warning. The heatmap theme comment now documents the intentionally black very-low segment. The focused utility suite passes: 55 tests.

src/Web/packages/ui/src/styles/nocturne-theme.css
smokkelaar, line 1: 2 times black, seems a error.
Also the comments are gone.
need to fix this

smokkelaar:
Addressed the latest browser-test review in b282cff: the page test now explicitly mocks its production remote/runtime/component imports, the previously dead panel stub is wired into those mocks, and all direct render-container queries were replaced with Vitest Browser locators plus narrowly scoped .element() access for layout/style measurements. The branch was also rebased onto current main. Local static diff checks pass; the local browser runner could not complete because its managed Chromium executable is not installed, so the browser suite still needs CI. The newly requested workflows currently show action_required before any jobs start and need a maintainer approval.

src/Web/packages/ui/src/styles/nocturne-theme.css
smokkelaar: Fixed in 78ff068f5. Only the exceptionally-low 40 mg/dL stop is black now; the following 54 mg/dL low stop is blue again, so adjacent ranges remain distinguishable. I also restored explicit inline meaning/color comments for both stops. The focused color utility suites pass: 55 tests.

src/Web/packages/app/src/routes/(authenticated)/reports/year-overview/year-color-focus.svelte.test.ts
```diff
@@ -0,0 +1,374 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { render } from "vitest-browser-svelte";
+import { page, userEvent } from "vitest/browser";
+import type { DailySummaryDay } from "$api/generated/nocturne-api-client";
+import { yearOverviewMocks } from "$lib/test-stubs/year-overview-remote";
+import {
+  page as applicationPage,
+  glucoseUnits,
+} from "$lib/test-stubs/year-overview-runtime.svelte";
```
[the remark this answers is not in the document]
smokkelaar: Resolved in 068a6726a. The page test now wires every production dependency it controls through explicit vi.mock calls: the generated Year Overview remotes, appearance store, date-parameter hook, YearHeatmap, filters, detail panel, and GRI chart. The previously unused EmptyYearPanel stub is now the replacement for the three nonessential panels, so it is no longer dead test code. The interaction tests also locate tracks through Vitest Browser page locators and use .element() only after locator resolution for measurements. This makes changes to yearOverviewMocks, applicationPage, and glucoseUnits affect the rendered page as intended.

ryceg:

---

**Requesting changes** — the feature is coherent and `metric-color-focus.ts` is genuinely well built, but it re-implements a preference system we already have, it carries an unrelated palette change that regresses dark mode, and the test suite that makes up half the diff is not run by anything.

First, credit where it is due, because most of this PR is solid. `resolveColorFocusRange` and `resolveGlucoseColorThresholds` reject every malformed shape I threw at them, `parseColorFocusPreferences` correctly drops bad entries without losing the good ones, and the piecewise-linear anchor remap in `glucoseColorFocusStops` stays monotonic for any input the validator accepts — I worked through a degenerate `[41, 42, 43, 44]` by hand to check. Making `getGlucoseHeatmapFill` take injectable stops is behaviour-preserving, since `GLUCOSE_HEATMAP_LEGEND_STOPS` is derived from `GLUCOSE_HEATMAP_STOPS`. And the `Number.isFinite` guards you added to `getMetricMax` and `getCellFill` are a real fix that stands on its own.

## Blocking

**1. Preferences should go through the store we already have, not a bespoke localStorage key.** `src/Web/packages/app/src/lib/stores/appearance-store.svelte.ts` already provides `SyncedPref`, which hydrates from the `nocturne-prefs` cookie before first paint, writes through to the backend user-preferences endpoint so a user's choices follow them across devices and tenants, and has a documented answer to the SSR shared-`$state` trap. `+page.svelte:76-155` instead hand-rolls a `localStorage` key namespaced by `JSON.stringify([tenantSlug, subjectId])`, its own parse and validate, its own load `$effect`, and its own storage-failure message — roughly ninety lines doing what the store does for free. `runed`'s `PersistedState` is already a dependency too, if a purely local option is wanted. You are importing `glucoseUnits` from that very store at `+page.svelte:26`, so it is right there.

The consequence is not only duplication. Because it is browser-local, two people looking at the same tenant's Year Overview see different colours, and a printed report will not match what the clinician renders from the same data. For a report that gets shared with a care team, that matters more than it would for an ordinary UI knob.

**2. The black heatmap stop is out of scope and regresses dark mode.** `nocturne-theme.css:17` changes `--glucose-heatmap-1` from `#2563eb` to `#000000`. There is exactly one definition of `--glucose-heatmap-*` in the file, in `:root` — the `.dark` block at line 115 only overrides the insulin, IOB and data-exclusion variables, so there is no dark-mode value for this. Empty cells render `rgb(0 0 0 / 5%)` over a dark card (`+page.svelte`, `getCellFill`), so in dark mode a black cell for an exceptionally low average becomes near-indistinguishable from a cell with no data at all. That is the opposite of what the new comment says it is for. It also drags the whole 40-54 mg/dL interpolation through black.

Your own PR description says the scope is "only the Year Overview presentation feature", and I agree with that scoping — please split this out. If we do want it, it needs a `.dark` counterpart and a look at the 40-54 ramp.

**3. Six new type errors.** I ran `svelte-check-native` on the merge base and on your head commit. Base `80b494084`: 44 errors across 34 files. Head `309c0b23e`: 50 across 37. The six new ones are yours:

- `ColorFocusRange.svelte:140` and `:155` — `Parameter 'event' implicitly has an 'any' type`
- `GlucoseColorThresholds.svelte:156` — same
- `year-color-focus.svelte.test.ts:66` and `:67` — `Type 'number | null' is not assignable to type 'number | undefined'` on `totalDailyDose` / `totalBolusUnits` in the `day()` factory
- `year-color-focus.svelte.test.ts:102` — unsound `as IntersectionObserverEntry` cast, needs `as unknown as`

The first three come from the `oninput={(event) => changeBound(...)}` handlers, because our shadcn `Input` wrapper does not type its event props. Annotating the parameter is enough. This is not your fault for missing it — `@nocturne/app`'s `check` is excluded from CI, so nothing told you.

**4. Nothing runs these tests, and the description's claim about CI is not correct.** The PR says "The repository CI remains authoritative for the full generated-client and browser suite", and your later comment says the browser suite "still needs CI". It will not get it. `.github/workflows/tests.yml`'s `web-typecheck` job runs vitest for `@nocturne/bot`, `@nocturne/portal` and `@nocturne/screenshots` only; `@nocturne/app` vitest, unit and browser both, runs in no workflow in this repository. There is a comment on that job explaining why `@nocturne/app` is excluded.

That matters because roughly 880 of the 1755 added lines are tests. Here is what I could verify myself:

- The node suites (`metric-color-focus.test.ts`, `chart-colors.test.ts`) pass — 55 of 55.
- The browser suites I could not validate. They fail on my machine with `effect_orphan` out of bits-ui's Slider, but so does the pre-existing `OidcProviderDialog.svelte.test.ts` (11 of 11) on the merge base, including after clearing `node_modules/.vite`. So that is my environment, not necessarily your code — but it does mean neither of us has seen these seventeen tests pass, and no automation will. Someone with a working browser-test setup needs to run them before this lands.

## Smaller things

**5. The two new components are near-duplicates.** `ColorFocusRange.svelte` (200 lines) and `GlucoseColorThresholds.svelte` (212) share the same skeleton throughout: `Slider.Root`, gradient span, the step-array workaround, draft inputs bound alongside an `invalidBound`, a reset button, duplicated screen and print blocks, and an identical `print-color-adjust` style rule. Only the thumb arity and the unit conversion actually differ. This repo leans hard on DRY and I would like one component here.

**6. `sliderSteps` is rebuilt on every pointer move.** `ColorFocusRange.svelte:30-42`. The 10,001-entry base array is correctly memoised — your first test pins that, which I appreciated — but the `Set` union and sort of about ten thousand numbers re-runs whenever `range` changes, which during a thumb drag is every `pointermove`. Sorting the base once and splicing in the two or three extra values would fix it.

**7. The storage-failure message will probably not be announced.** `+page.svelte:632`, `role={colorFocusStorageFailed ? "status" : undefined}`. A live region has to exist before its contents change; turning the role on at the same instant the text changes generally announces nothing. Make it `role="status"` unconditionally, or use a separate always-present region.

**8. `chart-colors.test.ts` now reads the theme CSS off disk** and asserts hex literals, in the test file for a different module. It will break on any retheme and it is testing a constant rather than behaviour. If the black stop survives point 2, assert it where the palette lives, or not at all.

**9. `void metricLabel` as the metric-identity signal.** `ColorFocusRange.svelte:47`. Using the display label to detect "we switched metric" works today but breaks silently the moment two metrics share a label. A `{#key selectedMetric}` around the component in `HeatmapLegend.svelte` deletes that `$effect` entirely and is robust.

**10. The average-glucose legend loses information.** The old SVG carried `role="img"` with an accessible name, the "Low / In Range / High / Very High" zone captions, and 70/180/250 tick labels. The replacement gradient is a bare `<span>` with no accessible name, and "In Range" is gone. The sliders do carry the numbers, so this is not a hard blocker, but the zone captions were doing real work for people scanning the report.

**11. Please think about how the threshold labels read.** Calling the four controls "Very low / Low / High / Very high" with defaults of 54/70/180/250 makes them look like clinical thresholds, and someone who sets "High" to 140 here may well believe they have changed their high target. Your "Color scale only; glucose targets and Time in Range are unchanged" note is a good mitigation and I am glad you anticipated it. Worth knowing that 250 is not actually a palette anchor by default — the anchors either side are 220 and 260 — so moving that handle remaps the whole ramp rather than shifting one boundary. Combined with point 1, a report's colours currently cannot be reproduced by anyone else, which is the part I would most like us to get right before this ships.

Thanks for splitting the Google Health and connector work out into a separate branch — that made this much easier to read.

smokkelaar:
Addressed the review in 7d85408af: preferences now use SyncedPref and the backend model (including validation, persisted resets, SSR isolation and shared appearance); the two controls are consolidated; exact slider steps no longer require re-sorting the base; metric changes use a key; and the legend again carries accessible zone descriptions and print values. Removed the CSS hex-literal test and the page's bespoke storage/error-message machinery.

One intentional departure from point 2: the feature requester explicitly wants a neutral endpoint. Their clarified requirement is a transition from 4 to 3 mmol/L, with a solid endpoint at/below 3. This is now black in light mode, white in dark mode, and black on white paper when printing. The color defaults are consequently 54/72/180/250 mg/dL, separate from glucose targets. All four themes were exercised in light, dark and print. Per-user sync is not described as forcing identical preferences on different signed-in users; the existing shared-appearance projection carries the owner's selection and the printed legend records the active scale.

Actual local validation: API/codegen build passed; 80 node/SSR tests, 30 browser tests and 47 focused backend tests passed. App check retains 34 existing diagnostics, with no new diagnostic relative to the reviewed base. The PR description now correctly states that app Vitest is not run by repository CI.

PR #1301, opened 2026-09-11
fix(oracle-cloud): converge the instance on every run, not just first boot

ryceg:
## The problem

The installer looked every cloud resource up before creating it, so re-running was safe on the OCI side. Everything *inside* the instance, though, was done by cloud-init — which runs exactly once, at first boot. A re-run hit `instance exists` and skipped straight past a broken server to the wait loop, so no failure on the guest could ever be repaired by running the script again.

Found on the first run against a real tenancy (this shipped unrun in #1259). The reserved public IP is attached *after* launch and there is no NAT gateway, so the instance boots with no route out and `curl -fsSL https://get.docker.com | sh` in `runcmd` failed. Later `runcmd` entries still ran, so `nocturne.service` was enabled — but it declared `Requires=docker.service`, a unit that does not exist until Docker is installed, so it sat `inactive (dead)` rather than `failed`. No diagnostic, and nothing the installer could fix on a second run.

## The change

All guest setup moves into an idempotent `/usr/local/sbin/nocturne-up`: firewall, Docker, release bundle, `.env`, DNS wait, `compose up`. cloud-init shrinks to planting the files and starting the unit — nothing in `runcmd` needs the network any more. The installer pushes the same files and runs the same script over SSH on **every** invocation, so a fix to `nocturne-up` reaches existing instances and an interrupted run resumes where it stopped.

Specifics:

- `nocturne-up` retries the Docker install and the bundle download rather than failing, so the boot-time run heals itself the moment the reserved IP is attached — the race above is no longer fatal.
- Dropped `Requires=docker.service`. `After=` alone is ignored while the unit is absent, so ordering still applies once Docker is there without the unsatisfiable dependency.
- `.env` converges: settings a newer release added to `default.env.example` are merged in, and only **blank** secrets are filled. Passwords are never rotated — regenerating one Postgres has already initialised itself with would lock Nocturne out of its own database.
- The bundle is re-downloaded when `NOCTURNE_VERSION` changes, so re-running with a newer tag actually upgrades instead of silently doing nothing.
- Repairs a route table that has rules but no default route (the old check only caught a completely empty table), and a security list missing any of 22/80/443 (it only checked 443). Both now test what they mean: the security-list check does proper range containment, so a deliberately wider rule is left alone.
- Starts a `STOPPED` instance and waits out transitional states.
- `--all` on the reserved-IP and instance lookups. Both paginate — the OCI CLI warns about it — and a truncated first page would have meant creating a duplicate.
- Hard-fails with a clear message if the generated cloud-init exceeds Oracle's 16 KB user-data cap. Currently 14,784 bytes encoded, so this is worth a guard.
- Folding@home setup runs from `nocturne-up` after egress is proven, instead of from `runcmd` where it had the same no-internet race.

## Docs

A tmux tip on the Oracle page — a dropped Cloud Shell session is exactly what leaves a run half-finished — and a callout that re-running is the fix for almost anything, which is now true rather than aspirational.

## Testing

No OCI tenancy in CI, so the cloud calls are still unexercised. What was checked:

- `bash -n` on the installer and on each guest script extracted back out of the rendered cloud-init.
- The generated cloud-init renders to valid YAML with all five `write_files` round-tripping intact, and the encoded size is measured against the 16 KB cap.
- The `.env` convergence logic exercised directly: fresh install; re-run leaves `INSTANCE_KEY` and every Postgres password byte-identical; a setting added to the example is merged in; a blanked secret is filled while a set one is preserved and a stale `BASE_DOMAIN` is corrected.
- The new route-table and security-list jq filters against nine hand-built inputs each (exact ports, a wide range covering all three, a range missing one, wrong protocol, `tcp-options: null`, stale gateway, empty).

## Verified against a live tenancy

The spend cap is no longer theoretical. Applied to a real Pay As You Go tenancy (`ap-melbourne-1`), the policy was accepted with all seven statements stored verbatim, and the effective limits read back as intended:

```
standard-a1-core-count        2.0
standard-a1-memory-count     12.0
standard-e4-core-count        0.0
```

The first two are the Always Free allowance exactly; the third is a paid shape, and `0` is the wildcard zero refusing it. Listing the real quota definitions is also what caught the regional-quota bug in the second commit — each Ampere shape has both a per-availability-domain and a regional quota, and setting only the per-domain pair back would have left the regional one at zero and blocked A1 launches region-wide.

The provisioning half of this branch is still unexercised end to end: no CI tenancy exists, so the VCN/instance/IP calls have not been run from this code. That is the same gap that let the original bug ship, and it has not closed.

### DIFF
--- a/deploy/oracle-cloud/oracle-cloud-install.sh
+++ b/deploy/oracle-cloud/oracle-cloud-install.sh
@@ -27,11 +27,13 @@
 #   BUDGET_EMAIL            where the spend alert goes (default: the Oracle account's email)
 #   COMPARTMENT_ID          where to create resources (default: tenancy root)
 #   NOCTURNE_VERSION        release tag to install (default: latest); re-run with this set to upgrade
-#   OCPUS / MEMORY_GB       A1 size (default 1 / 6; the free tier allows 4 / 24 in total)
+#   OCPUS / MEMORY_GB       A1 size (default 1 / 6; the free tier allows 2 / 12 in total)
 #   BOOT_VOLUME_GB          boot volume size (default 50; the free tier allows 200 in total)
 #   SSH_PUBLIC_KEY_FILE     key to authorise (default: generated at ~/.ssh/nocturne_oci)
 #   CAPACITY_RETRY_MINUTES  how long to keep retrying "out of host capacity" (default 30)
 #   DNS_WAIT_MINUTES        how long the instance waits for your DNS records (default 10)
+#   ALLOW_PAID=1            permit a size above the Always Free allowance, and skip the quota
+#                           policy that otherwise stops this tenancy creating anything billable
 
 set -euo pipefail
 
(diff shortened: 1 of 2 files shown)


ryceg:
Pushed a second commit adding spend enforcement, since the first run surfaced a live billing bug alongside the provisioning one.

Oracle [halved the Always Free Ampere allowance](https://www.infoq.com/news/2026/07/oracle-cloud-free-tier-limits/) from 4 OCPU / 24 GB to 2 OCPU / 12 GB on 15 June 2026 with no announcement, and terminated over-limit instances from 18 August. This script still advertised `the free tier allows 4 / 24` — and the docs page recommends upgrading to Pay As You Go to get around Ampere capacity errors, so anyone who followed both was being billed for the excess.

Three layers:

- Free-tier figures corrected, stated once, feeding both the size check and the quota policy.
- `OCPUS`/`MEMORY_GB`/`BOOT_VOLUME_GB` above the allowance fail before anything is created. `ALLOW_PAID=1` opts out.
- A tenancy-wide quota policy zeroes every compute core/memory quota except the Always Free shapes, so a paid resource is *refused* rather than billed. Budgets can't do this — they are alert-only on a 24-hour evaluation cycle.

Blast radius is deliberately limited: the policy is skipped when Nocturne is in a compartment rather than the tenancy root, when the tenancy already runs other instances, or when `ALLOW_PAID` is set. It is written after the instance is running, so a rejected policy can never block the launch it protects, and a creation failure is reported rather than fatal.

Caveat worth reviewing: compute quotas are scoped per availability domain, so in a multi-AD region this caps each domain rather than the tenancy total.

PR #1339, opened 2026-09-15
fix(connectors): store one canonical connector name

ryceg:
Production held two `connector_configurations` rows spelled `CareLink` and `carelink`, for different tenants. The PascalCase one was created 2026-09-12 and is actively polling, so this is a live write path, not historical drift.

## Mechanism

`ConnectorConfigurationService` stored the caller's raw spelling on write (`:169`, `:234`, `:455`) while every lookup lowercased (`c.ConnectorName.ToLower() == connectorNameLower`). Two writers disagreed:

- `ConfigurationController` passes the `{connectorName}` route segment through — the UI sends the lowercase id, which is why the other eleven connectors are lowercase.
- `CareLinkConnectController` declared its own `const string ConnectorName = "CareLink"` and called `SaveSecretsAsync` with it, which **creates** the row before any configuration write.

CareLink is the only connector with a bespoke connect controller — headless password login is CAPTCHA-blocked and Medtronic's Auth0 client only allows the `com.medtronic.carepartner:/sso` redirect, so the paste-the-code flow is the only onboarding route, and it never touches the settings PUT that would have supplied the lowercase name. The capital L is a red herring; `ConnectorRegistrationAttribute.ConnectorId` already lowercased correctly.

## What the split actually costs

Lookups had already been hardened piecemeal (the token cache carries a comment naming this exact drift), so the casing alone breaks no read today. What it does cause:

1. The unique index is over `(connector_name, tenant_id)` and is **case-sensitive**, so it does not enforce one config per connector per tenant — two writes on different casings both insert.
2. `GetAllConnectorStatusAsync` builds an `OrdinalIgnoreCase` dictionary, so a tenant holding both casings throws a duplicate-key `ArgumentException` and **500s the whole connectors settings page**.
3. `lower(connector_name) = $1` could not use the index, so every connector-config read was a scan.

## Fix

Canonical form is lowercase — confirmed rather than assumed: `ConnectorRegistrationAttribute.ConnectorId` documents itself as "the stored `ConnectorConfiguration.ConnectorName` lowered", and it is both the API route segment and the UI's id.

The chokepoint is the entity that owns the column: `ConnectorNames.Canonical` is the single rule and `ConnectorConfigurationEntity.ConnectorName`'s setter applies it, so no write path can store a second spelling. Lookups become `==` and can use the index. A `ConnectorName` value type was considered and rejected — it would ripple into MVC route binding, the OpenAPI surface and the generated TS client for no safety the setter does not already give.

Duplicate name sources removed: CareLink's controller constant now derives from the registration, and the ten `protected override string ConnectorName => "…"` declarations collapse into one `NameFor(typeof(TConfig))` default.

**That collapse fixes a second live bug.** `LibreLinkAuthTokenProvider` declared `"FreeStyle"` while its config registers as `"LibreLinkUp"`. The token cache keyed sessions under `freestyle`, but `InvalidateCaches` passes the lowercase route id `librelinkup` on a secret save — so **a LibreLinkUp user changing a wrong password never evicted the cached session**, and kept using the stale token until it expired.

## Migration

`NormalizeConnectorNameCasing` merges duplicates key-wise (winner = the row that most recently proved it could sync, its keys taking precedence), deletes the losers, lowercases, then adds `CHECK (connector_name = lower(connector_name))` — all in one transaction, constraint last.

The CHECK rather than a functional unique index: once no second spelling can exist, the existing `(connector_name, tenant_id)` index already means one row per connector per tenant, so a `lower()` index would be redundant. And the CHECK closes the failure the index cannot — a single mis-cased row is *invisible* rather than duplicated, and only the CHECK rejects it. An old instance mid-rolling-deploy gets a loud `23514` instead of silently recreating the split.

Merging folds `configuration` and `secrets` only; a loser's `sync_cursors` and `backfill_low_water_marks` go with the deleted row. That costs a re-crawl, not data — the winner is by construction the most recently synced row, so it holds the more advanced cursors anyway.

`jsonb_each` raises on a non-object, which would abort the block and leave the API unable to start, so both arguments degrade a non-object to `'{}'`.

## Verification

Postgres 17 harness with the real unique index, RLS policy and FORCE RLS, run as both a superuser and a `NOSUPERUSER NOBYPASSRLS` role. Seeds: both casings for one tenant, three casings, a PascalCase row holding the recent successful sync, non-object jsonb, untouched sibling connectors, and a tenant with only one spelling. Idempotent on re-run and a no-op on already-normalised data.

That harness caught a real bug in the first draft: the statements relied on RLS alone for tenant scoping, so under a role that bypasses RLS the ranking ran across **all** tenants and deleted other tenants' live rows. Every statement now also restricts on `tenant_id`; mutating that filter back out deletes 10 of 15 seeded rows as a superuser and nothing as the RLS-bound migrator.

Also `tests/Integration/…/ConnectorNameCasingTests.cs` (Testcontainers): migrate to the prior migration, seed, migrate to head, assert the rename, the merge, an untouched sibling, and both constraint rejections. 6/6.

Dry-run against the current production table: **0 deletions, 1 rename.**

`Infrastructure.Data` 1043, `Nocturne.API.Tests` 6850 passed / 5 skipped.

## Note for whoever merges

Conflicts with #1334 on `LibreLinkAuthTokenProvider`'s name — that PR changes the literal to `"LibreLinkUp"` to fix the same bug standalone, this one removes the override entirely in favour of the registration-derived default. **This branch's version is the better resolution.**

Also worth knowing: this renames the live CareLink row and closes the duplicate hazard, but will **not** make that tenant's connector sync — its `"Authentication failed"` is a genuine Medtronic auth failure needing the refresh token re-established through the connect flow.


Uneo7:
tested, gtg

Issue #608, opened 2026-07-29
Connection error on top right corner [connected to Trio]


railing236:
![Image](https://github.com/user-attachments/assets/ce2ffce1-7ca7-458a-ab07-a7e67a3efacb)

Despite the connection being active and correct (I see all data coming from Trio and the data updates every 5 min), I get connection error message which is a bit annoying really.

PR #1348, opened 2026-09-17
fix(web): let realtime fall back to polling, and stop reporting dead sockets as connected

ryceg:
Users report **"Connection Error"** on the live reading while their data keeps updating. Three separate defects in the browser's realtime path produce that symptom. Closes the same complaint as #608.

## The socket was effectively WebSocket-only

The client passes `transports: ["websocket", "polling"]`, but engine.io-client 6.6.4 defaults `tryAllTransports` to `false` — a failed first transport ends the attempt instead of trying the next, so the `"polling"` entry was unreachable:

```js
// engine.io-client/build/esm/socket.js
_onError(err) {
  if (this.opts.tryAllTransports && this.transports.length > 1 && this.readyState === "opening") {
    this.transports.shift();
    return this._open();
  }
  this.emitReserved("error", err);   // <- we always landed here
```

Any network that blocks WebSocket (corporate proxies, some mobile carriers and VPNs, privacy extensions) left realtime in a permanent `connect_error`, while REST backfill quietly kept the dashboard current — which is what makes the fault so confusing to report. The bridge serves both transports, so falling back keeps those users live.

## A ticket-less handshake was reported as a healthy connection

When `/realtime/ticket` returns no token, the client handed the bridge `token: ""`. The bridge reads an empty token as *"no ticket offered"* and admits the socket down the legacy Nightscout path — which joins **no tenant room**. The socket really was open and really received nothing, while the UI showed "connected".

Verified against production: the ticket endpoint returned a `null` token on one run and minted fine on the next, with latency ranging 0.8s–8.4s against the client's own 8s abort — so this is reachable in practice, not theoretical.

Such a handshake is now dropped rather than left masquerading as working: a transient ticket failure retries with backoff, and a definitive denial reports a new `unauthorized` status.

## Only the sidebar debounced the disconnected state

The dashboard, mobile header and command palette flipped to "Connection Error" the moment `connectionStatus !== "connected"` — including the ordinary connect on page load and every reconnect backoff. That logic now lives in one place (`connection-indicator.svelte.ts`) and treats `connecting`, `reconnecting` and `unauthorized` as nothing to report.

## Tests

10 new tests, each confirmed to fail with its fix reverted:

- `socket-options.test.ts` stands up a WebSocket-hostile HTTP server (answers HTTP, destroys the upgrade) and asserts the engine actually reaches the polling transport. With `tryAllTransports: false` it observes only `['websocket']`.
- `websocket-client.test.ts` replays a handshake with and without a ticket against a fake socket, covering connected / dropped / retried / unauthorized.
- `connection-indicator.test.ts` pins which statuses count as an error.

Typecheck error count is unchanged from baseline (1175 both with and without this change — the tree cannot typecheck clean without API codegen), and no new lint problems.


Uneo7:
tested, gtg

PR #1351, opened 2026-09-17
fix(connectors): tell a Nightscout user their site was unreachable, not that their secret was wrong

### LINKED ISSUE #1220: Nightscout connector reports a transport failure as "Authentication failed"
## Summary

The Nightscout connector tells a user their credentials are wrong when the real problem is that Nocturne cannot reach their site. Verified against main at d3b107c2a.

`NightscoutConnectorService.AuthenticateAsync` (`src/Connectors/Nocturne.Connectors.Nightscout/Services/NightscoutConnectorService.cs:~120-127`) wraps the whole status probe in `catch (Exception ex)` and returns `false`, so an `HttpRequestException` (DNS failure, connection refused, silently dropped packets) or a `TaskCanceledException` (timeout) is indistinguishable from a 401. `BaseConnectorService` (`src/Connectors/Nocturne.Connectors.Core/Services/BaseConnectorService.cs:~1293-1296`) turns any `false` into `AuthenticationFailedResult()`, whose user-facing error is `Authentication failed for nightscout-connector` (`:~143`).

This is the connector half of the real-world report behind #1185: the user's Nightscout was dropping packets from the server's IP, the secret was never sent, and the connector told them to fix their API secret.

## Fix

Distinguish transport failures from a real 401/403 response in `AuthenticateAsync` and carry the cause out, either through a richer return than `bool` or by setting the connector's error message before returning, so the base class stops overwriting it with the authentication wording. The #1185 fix (PR pending) introduces the plain-language message set for the migration path ("Could not reach your Nightscout server. Check it is online and that it allows connections from Nocturne." / "Nightscout rejected the API secret…"); reuse those strings rather than writing a second set. The connector already special-cases Cloudflare challenge pages; a silent TCP drop needs the same care.

Check whether the other nine connectors' `AuthenticateAsync` overrides have the same shape before fixing only this one; if they do, the fix belongs in the base's handling of the auth result.

Split from #1185 (maintainer comment, bug 2).

ryceg:
Closes #1220.

## The bug

`NightscoutConnectorService`'s auth probe wrapped everything in one `catch` and returned `false`. `BaseConnectorService` turned every `false` into a fixed "Authentication failed", so a DNS failure, a refused connection, dropped packets or a client timeout sent the user to check an API secret that was never at fault. The same discard lost the connector's Cloudflare/WAF explanation: it was written, passed to `TrackFailedRequest`, and logged into the void. The tenant never saw it.

## The fix

- **The probe classifies.** A site that never answered (`HttpRequestException`, client timeout) reports the reachability wording; `OutboundRefusedException` passes its own message through; a 401 reports the rejected-secret wording; a 403 reports a refusal without naming the secret; anything else reports a fixed sentence and leaves the exception text in the log. A cancellation the caller asked for propagates instead of being reported as a failed credential, and the caller's token now reaches the HTTP call.
- **One set of strings, one selector.** `NightscoutMessages` (Core.Constants) holds the plain-language strings the setup-time migration already used, plus `ForStatus(...)`, which is the only place a status becomes a message. `MigrationJobService` now reads through it, so the import and the connector cannot describe the same failure differently.
- **401 and 403 are no longer the same answer.** Nightscout sends 401 for a secret it does not recognise. A 403 is as likely an IP rule, proxy or non-Cloudflare WAF in front of the site, so naming the API secret there would recreate this bug for everyone it turns away. A 404 on either probe (both ask for a route every Nightscout serves) says the address is not a Nightscout; the import's collection walk keeps reporting a 404 as that collection's status, since an older site may lack one. A 5xx says the site is unwell and to try again, keeping the number for support. The migration keeps its `ApiSecretRejected` cause for 403 (its subjects step relies on it to skip a non-admin secret) but no longer borrows the secret wording.
- **`TrackFailedAuthentication(reason)`** on the base records a tenant-facing reason that `AuthenticationFailedResult()` reports in place of the generic wording. It is deliberately separate from `TrackFailedRequest`, whose reasons are log-grade ("HTTP Unauthorized", "JSON parsing error") and must not become what a tenant is told to do. Opt-in, so no other connector's output changes in this PR.

Only Nightscout overrides `AuthenticateAsync`/`EnsureAuthenticatedAsync`; the other connectors authenticate inside their own sync bodies and call `AuthenticationFailedResult()` directly, so they still show the generic wording. Follow-up: #1352.

## Tests

- 12 new Nightscout facts: unreachable (`HttpRequestException` + `SocketException`), client timeout, 401, 403, 404 and 500 wording asserted literally (no mention of the secret), Cloudflare challenge page, `OutboundRefusedException` pass-through, unexpected exception kept out of the tenant's text, blank secret asks for the secret, caller-cancelled token propagates, both entry points.
- 2 base facts in `SyncTerminalPhaseTests`: a recorded reason reaches `SyncResult.Message`, `Errors` and the terminal progress event on both entry points.
- Migration tests: a 403 is a refusal and not a bare status; the connection test reads a 404 as a wrong address while a collection 404 keeps its status wording; the 401 wording offers a blank secret only on the import.

| Suite | Result |
|---|---|
| Connectors.Nightscout.Tests | 82 passed |
| Connectors.Core.Tests | 221 passed |
| CareLink / MyFitnessPal / Tidepool / NocturneRemote | 99 / 52 / 14 / 28 passed |
| API.Tests (Migration filter) | 54 passed |

Not run against a live Nightscout; transport paths are exercised through faked handlers.


Uneo7:
Tested, gtg.

Issue #631, opened 2026-08-02
Feature request: Docker volume bind option


achkars:
A nice to have feature would be a clearly labeled configurable Docker bind location for the nocturne-postgres-data volume in the compose file or env vars.

Would save users that will run Nocturne service locally from having to play around with compose files.

Thanks!

smokkelaar:
Can you explain more? @achkars ?

achkars:
By default, the compose file stores DB data in the nocturne-postgres-data container (Mount path: /var/lib/docker/volumes/nocturne_nocturne-postgres-data/_data).

I would like to have a environment variable to be able have the DB data saved in an alternate location to have the data saved outside the default container volume.

This would allow easier backups for people running various docker containers.

smokkelaar:
@achkars Can you check #1355  is this is the solotion that is needed?

I'm not sure myself. and it needs to be tested before i make it a pull request.

achkars:
That looks right. I can give it a try.

PR #1355, opened 2026-09-17
feat(deploy): configurable Postgres data mount path

smokkelaar:
## Summary

Adds an optional, backwards-compatible way to redirect the Postgres data directory to a host path of the operator's choosing, instead of Docker's default managed-volume location.

## What changed

The Postgres data mount (`/var/lib/postgresql/data`) in `deploy/docker-compose/docker-compose.yaml` and `deploy/portainer/docker-compose.yaml` now reads its `type` and `source` from two optional environment variables:

```yaml
volumes:
- type: ${POSTGRES_DATA_MOUNT_TYPE:-volume}
  target: /var/lib/postgresql/data
  source: ${POSTGRES_DATA_PATH:-nocturne-postgres-data}
  read_only: false
```

Left unset, both variables resolve to today's defaults (`volume` / `nocturne-postgres-data`) — byte-for-byte the same effective config as before this change. No new required input, and no change to the existing install flow.

To opt in, an operator sets both variables, e.g.:

```env
POSTGRES_DATA_MOUNT_TYPE=bind
POSTGRES_DATA_PATH=/srv/nocturne/postgres-data
```

The source-of-truth change is in `src/Aspire/Nocturne.Aspire.Host/Program.cs` (the Aspire AppHost that generates these two compose files via `scripts/publish-release.cs`); the committed `deploy/` YAML is updated to match so the change is usable immediately, not only after the next generated release.

## Out of scope (intentionally)

- No documentation changes yet (README/DEVELOPMENT.md, `.env.example` entries) — planned as a follow-up.
- No migration tooling for existing installs that want to move data from the current named volume to a bind path; that remains a manual stop/copy/chown/start operation.
- Not linked to any existing issue.

## Validation

- `dotnet build src/Aspire/Nocturne.Aspire.Host/Nocturne.Aspire.Host.csproj` — succeeds.
- Both edited compose files parse as valid YAML and resolve to the exact previous default values when the new variables are unset.


smokkelaar:
https://github.com/nightscout/nocturne/issues/631
@ryceg  This needs a very carefull review. since i cannot test this by myself.

Uneo7 (changes requested):
Hey,

 Thanks for the PR, I reviewed it and ran it locally.

What I checked:

- `aspire publish`: the generated files match what you committed.
- Both variables set: Postgres starts on the host path, roles created, data survives a restart.
- Neither set: renders exactly as today.
- Type only, on an install that already held a row: table gone after the restart.

And I found two issues : 

First : nothing ties the type and the source together, so setting only one of the two is easy, and the two halves fail very differently:

1. `POSTGRES_DATA_MOUNT_TYPE=bind` alone: Compose reads `nocturne-postgres-data` as a relative path, creates `./nocturne-postgres-data`, and initdb runs into it. No error, but the site looks wiped.
2. `POSTGRES_DATA_PATH` alone: Compose refuses the project and nothing starts, fine.

Can we use an optional override instead, and leave `docker-compose.yaml` alone?

```yaml
# deploy/docker-compose/docker-compose.bind-data.yaml
services:
  nocturne-postgres-server:
    volumes:
    - type: "bind"
      target: "/var/lib/postgresql/data"
      source: "${POSTGRES_DATA_PATH:?set POSTGRES_DATA_PATH to the host directory that should hold the database}"
      read_only: false
```

One required variable, fixed type, so the two halves cannot disagree:

```bash
# default install, unchanged
docker compose up -d

# data in a host directory
POSTGRES_DATA_PATH=/srv/nocturne/pg \
  docker compose -f docker-compose.yaml -f docker-compose.bind-data.yaml up -d
```
With this, if you forget the variable, Compose will stop before any container starts, so initdb never touches the wrong directory.

`ByoProxyComposePublisher.cs` already ships an override this way, worth a look for the publisher side.

Second one :  `Program.cs:185` still uses the hardcoded data something like `$"${{POSTGRES_DATA_PATH:-{ServiceNames.Volumes.PostgresData}}}"` should fix it.

Finally could you update the doc as well, that way it's doen't get forgotten :)

Thanks

PR #1361, opened 2026-09-19
fix(web): show lab results in eHbA1c tooltips

### LINKED ISSUE #1360: eHbA1c lab result data points
## Description

When adding an HbA1c lab result on the eHbA1c screen, the lab result appears in the chart, but when you hover the mouse over the black triangle, the label that appears still shows the purple dot, meaning an eHbA1c value, not the black triangle, meaning a lab result. The proper value for the lab result is shown in the hover text, but the legend is incorrect when hovering over the lab result.

## Steps to Reproduce

1. Open the Reports section of the left navigation pane.
2. Add an HbA1c lab result using the data entry fields beneath the chart.
3. Locate the black diamond added to the chart above, and hover your mouse over the black diamond.
4. Look at the hover text that pops up over the reading. The value you entered is shown, but the legend shows a purple dot representing the eHbA1c line, and is labeled "eHbA1c".

## Expected Behavior

The legend should show a black triangle to match the one shown in the chart, and it should be labeled as "lab result", not "eHbA1c."

## Actual Behavior

The legend in the hover text box shows a purple dot which is labeled as "eHbA1c".

**Screenshots:** 2 attached

<details>
<summary>Diagnostic Info</summary>

```json
{
  "userAgent": "Mozilla/5.0 (X11; Linux x86_64; rv:151.0) Gecko/20100101 Firefox/151.0",
  "screenSize": "1470x1152",
  "route": "/settings/support",
  "locale": "en-US"
}
```

</details>

Image
Image

smokkelaar:
Fixes #1360`n`nShow the lab result values in the eHbA1c report tooltips, including the existing eHbA1c tooltip changes only.


smokkelaar:
Resolved the tooltip review comment in 681631d. Lab-only dates are now included in the chart x-domain with a null y-value, so LayerChart can resolve and hover those dates without drawing a fake eHbA1c value. Dates that have both values still render the combined tooltip; dates with only one value show only that value. This preserves the existing calculation and marker behavior while removing the dependency on an eHbA1c point for lab-date tooltip lookup.

smokkelaar:
Follow-up fix in 0d6ccb0: multiple lab results on the same calendar date are now collected with filter() and rendered as separate rows, so no same-day result is dropped. Test C c12 (HA wrapper 6ba73f0) validates this on the official Daily/main runtime basis.

smokkelaar:
Follow-up fix in dbc4154: lab-only dates remain in the tooltip x-domain, but now receive a linear display-only interpolation between neighboring real eHbA1c points. This keeps the visible line continuous without presenting the interpolated value as a measured eHbA1c value. Test C c13 (HA wrapper 5fdf898) validates the result on the official Daily/main runtime basis.

smokkelaar:
> ## Copilot review overview
> ### 🔵 Needs a closer look
> Date normalization can show the wrong day, while synthetic tooltip points can alter or extend the plotted eHbA1c curve.
> 
> **Review effort:** Balanced **Findings:** None
> 
> **Previously missed (1)**

Since it is a very slow process, is theoretical 1 day difference not important.
Atleast not for this Bug ticket.

Uneo7:
Thanks for this.

I had a version of this that only replaced the tooltip snippet and left the series data alone. Yours is better: it shows both the eHbA1c and the lab row when they fall on the same date, and it carries the note. Mine showed only one of the two, picked arbitrarily by the quadtree.

Two things before merging:

1. Could you add a test? Hovering a lab marker should name it "Lab result", and hovering the line should still say "eHbA1c". There's a pattern to copy in `compression-lows-page.svelte.test.ts`.

2. `displayChartData` now gets a synthetic row per lab date with an interpolated value, just to give the hover a target. It works, and the summaries are unaffected since `extremes`/`yDomain` read `chartData`, but it means the series array now contains rows that exist because a lab result exists. Worth a comment saying so, given the note right above `labChartPoints` about lab draws never feeding back in.

One thing you may not have hit: declaring the lab draws as a second series would have been the obvious fix, and it breaks, `LineChart` renders a `Spline` per visible series, so the lab points get connected by a line. You avoided that, just noting it so nobody "simplifies" it later.

PR #1363, opened 2026-09-19
Avoid false connection errors after tab resume

### LINKED ISSUE #1362: Connection Error shown after returning to a backgrounded browser tab
## Summary

When Nocturne is left in a background browser tab, Chrome-family browsers may suspend or throttle the tab and the Socket.IO transport can disconnect. Returning to the tab briefly shows **Connection Error**, even though the service is healthy and current data is restored a few seconds later.

The message represents a transient browser transport state rather than an actionable service outage, so it can unnecessarily alarm users.

## Steps to reproduce

1. Open an authenticated Nocturne dashboard in Chrome or Dia.
2. Leave the tab in the background long enough for browser throttling or suspension to affect the realtime connection.
3. Return to the Nocturne tab.
4. Observe **Connection Error** while the socket reconnects and the existing visibility-triggered backfill refreshes the data.

## Actual behaviour

Several UI surfaces derive their error state directly from `!realtimeStore.isConnected`. A momentary Socket.IO disconnect is therefore presented immediately as a user-facing connection failure. The store also emits a connection error toast immediately on `connect_error`.

The existing visibility handler performs a forced backfill when the page becomes visible, but it does not explicitly ask a disconnected socket to reconnect.

## Expected behaviour

- Background-tab transport interruptions should not be presented as service outages.
- Returning to a visible tab should explicitly ensure that the realtime socket is reconnecting and should continue to backfill missed data.
- A user-facing warning should appear only when the connection remains unavailable for a meaningful period while the page is visible.
- All connection indicators should use one consistent presentation policy.

## Proposed fix

- Add an idempotent `ensureConnected()` operation to the websocket client.
- On `visibilitychange` to visible, clear any stale warning, ensure the socket is connecting, and retain the forced backfill.
- Suppress connection warnings while the document is hidden.
- Delay the public unavailable state for 10 seconds while visible, clearing it immediately on reconnect.
- Make dashboard, header, sidebar, command palette, and layout indicators consume the debounced unavailable state rather than the raw socket state.
- Add tests covering sustained, transient, and hidden-page disconnects.

AndyLow91:
## Summary

- separate the raw Socket.IO state from the user-facing realtime availability state
- suppress connection warnings while the page is hidden and allow 10 seconds for a visible-page reconnect
- explicitly resume the socket when a tab becomes visible, while retaining the existing forced backfill
- use the same debounced availability state across the dashboard, mobile header, sidebar, command palette, and authenticated layout
- add regression coverage for sustained, transient, and hidden-page disconnects

This prevents browser tab suspension from being presented as an application outage while still surfacing a connection that remains unavailable when the user is actively viewing the page.

Closes #1362

## Validation

- `vitest --config vitest.config.ts run src/lib/stores/realtime-store.test.ts` — 19 tests passed
- production web image built successfully with `Dockerfile.web`
- `git diff --check`


Uneo7:
Hey Andy,

Thanks for the chages. I ran your tests, then wrote a throwaway one driving the `visibilitychange` listener directly, and that one turned up two problems. Both are in the visibility handler, worth fixing before merging.

1. Line 325 schedules a notice without checking anything first, so users who aren't allowed realtime (denied ticket, or no INSTANCE_KEY) get warned 10s after every tab focus, even though the client deliberately keeps quiet for them. Guarding it with `if (!this.websocketClient.isConnected)` covers most of it.

2. Lines 322, 331 and 333 reset connection state that belongs to the `connect` handler. 333 makes the toast come back on every tab switch during a real outage (two hide/show cycles, two toasts), and clearing `connectionUnavailable` shows a healthy badge for 10s when you return to a connection that died 20 minutes ago. Dropping the three assignments and keeping only `clearDisconnectNotice()` on both branches should do it. Your case still works, a socket that dies while hidden never starts the timer anyway.

One open question: should a socket that never connected warn at all? A denied ticket is meant to stay quiet, but a first connect that fails because the network was down is the one you do want to show.

Could you keep a test on that path? The current three only cover socket events, which is why none of this showed up.

Minor:
- `ensureConnected()` doesn't call `clearAuthRetry()`, so a pending auth retry runs alongside the manual connect.
- Leftover `// Connection status` comment in CurrentBGDisplay.svelte:47.
- `BgDeltaWidget.svelte:44` still uses raw `isConnected`, so it flips to WifiOff on resume. Deliberate?

Uneo7 (changes requested):
See previous comment

PR #1380, opened 2026-09-22
fix(realtime): isolate Home Assistant glucose relay

### LINKED ISSUE #1379: bug(realtime): DataHub failure suppresses Home Assistant glucose relay
## Summary

`SignalRBroadcastService.BroadcastDataUpdateAsync` sends a normal `dataUpdate` to `DataHub` and then relays the same reading to `HomeAssistantHub`. Both sends currently share one `try` block.

If the `DataHub` send throws, control jumps directly to the catch block and the independent Home Assistant `glucose_reading` relay is skipped, even when Home Assistant is still reachable.

## Impact

A transient failure in the regular SignalR data channel can suppress a successfully persisted glucose reading from connected Home Assistant subscribers. The overview relay is already isolated in its own best-effort block; the Home Assistant relay is not.

## Reproduction

On current `main` (`bfb0786c56b285457add9280b656b5ab9b369b8b`):

1. Configure the `DataHub` group proxy to throw from `SendCoreAsync("dataUpdate", ...)`.
2. Keep the `HomeAssistantHub` group proxy operational.
3. Call `BroadcastDataUpdateAsync`.
4. Observe that `SendCoreAsync("glucose_reading", ...)` is never invoked on the Home Assistant proxy.

A focused regression test observes 0 Home Assistant calls where exactly 1 is expected.

## Expected behavior

The `DataHub`, `HomeAssistantHub`, and `OverviewHub` broadcasts are independent best-effort targets. Failure of one target should be logged without suppressing the remaining targets.

## Actual behavior

A `DataHub` exception suppresses the Home Assistant relay because both calls share one `try` block.

## Suggested fix

Move `BroadcastHomeAssistantGlucoseAsync(data)` after the `DataHub` catch. That helper already catches and logs its own failures, so this preserves write-path best-effort semantics while isolating the two SignalR destinations.

## Duplicate search

Searched open and closed issues and pull requests for combinations of:

- `dataUpdate` + `Home Assistant`
- `glucose_reading` + `dataUpdate`
- `broadcast failure` + `Home Assistant`
- `SignalR HomeAssistant relay failure`

No matching report or fix was found. Merged PR #216 establishes the intended Home Assistant glucose relay but does not address this failure isolation case.

BlackSoul697:
## Summary

- isolate the Home Assistant glucose relay from failures in the normal `DataHub` broadcast
- add a regression test proving a `DataHub` exception no longer suppresses `glucose_reading`
- preserve the existing best-effort behavior and independent overview notification

Fixes #1379

## Root cause

`BroadcastDataUpdateAsync` performed the regular `dataUpdate` send and the Home Assistant relay inside the same `try` block. When the first send threw, control skipped the Home Assistant call entirely, even though its hub could still be available.

`BroadcastHomeAssistantGlucoseAsync` already owns its own exception handling, so invoking it after the `DataHub` catch makes the two destinations independent without changing write-path behavior.

### Regression proof

With the regression test present but the production change temporarily reverted:

```text
Failed: 1 / 1
Expected glucose_reading once; observed 0
```

After restoring the fix:

```text
SignalRBroadcastServiceTests + TenantIsolationTests
Passed: 69 / 69
```

Also verified:

- `git diff --check`
- only the service and its focused unit-test file changed
- current upstream base: `bfb0786c56b285457add9280b656b5ab9b369b8b`

## Risk

Low. The change only separates two existing best-effort broadcasts. Home Assistant failures remain caught and logged by `BroadcastHomeAssistantGlucoseAsync`.


Uneo7:
Hey,

Thanks for the fix, I added a bunch of test. When possible please add the appropriate tests it helps with the review :)

